PR #2, opened 2018-06-27
Adds barebones pom.xml

cburns93:
- Adds initial pom.xml for the skeleton of the projects. This includes, the 3 build profiles (dev, dev-publish, release-publish), the correct maven plugins for compiling, creating jars and also the junit testing dependencies

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,180 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>company-accounts.api.ch.gov.uk</artifactId>
+    <version>unversioned</version>
+    <packaging>jar</packaging>
+    <name>company-accounts-api</name>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -0,0 +1,180 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>uk.gov.companieshouse</groupId>
+    <artifactId>company-accounts.api.ch.gov.uk</artifactId>
+    <version>unversioned</version>
+    <packaging>jar</packaging>
+    <name>company-accounts-api</name>
+    <description>Company accounts API to handle company accounts</description>
```
drosser: Maybe say "to handle company accounts data" (or something along those lines - makes it a bit clearer)

pom.xml
```diff
@@ -0,0 +1,180 @@
...
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>${mockito-junit-jupiter.version}</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <profiles>
+        <!-- release-publish is the profile that is activated when the build server builds and
+             publishes a release project to the cloud artifactory -->
```
drosser: Do these comments need to be here (applies to the comments on the other profiles too)

cburns93: Yeh, I put these there as people where asking what the difference between the profiles were. This hopefully clarifies it a bit

pom.xml
cburns93: Done

PR #6, opened 2018-07-06
Implements creation of accounts resource

cburns93:
- Adds the persistence entities and rest models
- Adds `java.time.LocalDate` module to the project to enforce the new Java 8 `LocalDate` library
- Adds the accounts repository functionality for use by Spring Data
- Adds the accounts service implementations for the creation of the accounts resource

Resolves:
SFA-337


src/main/java/uk/gov/companieshouse/api/accounts/controller/AccountController.java
```diff
@@ -0,0 +1,24 @@
...
+import uk.gov.companieshouse.api.accounts.model.rest.Account;
+import uk.gov.companieshouse.api.accounts.service.AccountService;
+
+@RestController
+public class AccountController {
+
+    @Autowired
+    private AccountService accountService;
+
+    @PostMapping(value = "/transactions/{transactionId}/accounts",
```
jsimons1989: There is current discussion about changing this to company-accounts instead of accounts to simplify migration to the new repo. Are you going to raise that change as a change request to update it once the decision is made?

cburns93: ummm, if its merged by then then yes i will raise a new pr

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private String etag;
+
+    protected AccountDataEntity build(String id){
+        this.id = id;
+        this.etag = GenerateEtagUtil.generateEtag();
+        this.kind = Kind.ACCOUNT.getValue();
+
+        Map<String, String> links = new HashMap<>();
+        links.put(LinkType.SELF.getLink(), "links/accounts");
```
jsimons1989: Why has this been hardcoded to "links/accounts"?

cburns93: It's a default value current as we dont know what the format of the links will be - the actual implementation will depend alot of the discussion with the architects

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
```diff
@@ -0,0 +1,36 @@
...
+            AccountEntity accountEntity = new AccountEntity().build(account);
+
+            AccountEntity createdAccount = accountRepository.insert(accountEntity);
+            if (createdAccount != null) {
+                return ResponseEntity.status(HttpStatus.CREATED).body(createdAccount.getData());
+            } else {
+                return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+            }
+        } catch (DataAccessException dataAccessException) {
+            return new ResponseEntity(HttpStatus.INTERNAL_SERVER_ERROR);
```
jsimons1989: The service should not be setting HTTP response codes. It should only be handling the logic. The controller has knowledge of HTTP requests and responses.

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
jsimons1989: We shouldn't be hardcoding temp values. They could get missed and accidentally released.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
cburns93: What would you recommend as a change?

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
cburns93: It won't get missed, it is the thing Steve is working on for the Small-full accounts PR. This will be the first thing that we do when that discussion has been had.

jsimons1989: It should be removed and added when Steve implements his part.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
jsimons1989: Return a boolean for whether the creation was successful from the service and set the appropriate response header in the controller. The exception/failure reason can be logged from the service.

cburns93: How would we then return the data of the account?

jsimons1989: Sorry, didn't spot the entire account being returned. You could return the account from the service and check if it is null in the controller for setting the HTTP codes.

The returned account from the service won't contain the etag, kind and links because you are setting them in the build method in the DB entity?

cburns93: So you want it to return the rest object instead?

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
cburns93: done

pom.xml
```diff
@@ -6,15 +6,15 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>uk.gov.companieshouse</groupId>
     <artifactId>company-accounts.api.ch.gov.uk</artifactId>
-    <version>0db2b2f</version>
+    <version>9857df7</version>
```
drosser: Should this be unversioned?

pom.xml
```diff
@@ -107,6 +107,10 @@
             <version>${mockito-junit-jupiter.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
```
drosser: Should we version this?

src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.companieshouse.api.accounts;
+
+public enum Kind {
+
+    ACCOUNT("accounts"),
+    ABRIDGED_ACCOUNT("accounts#abridged"),
```
drosser: Should we be supporting abridged accounts at this time?

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
```diff
@@ -0,0 +1,79 @@
...
+        this.kind = Kind.ACCOUNT.getValue();
+
+        Map<String, String> links = new HashMap<>();
+        
+        links.put(LinkType.SELF.getLink(), "");
+        this.links = links;
+        return this;
+    }
+
+    public String getId() {
```
drosser: Can we not use something like Project Lombok to clean up the get/set bloat?

pom.xml
cburns93: done

src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
cburns93: Nothing wrong with putting it in, besides, this may go with Steves code after the refactor

pom.xml
cburns93: Was thinking about it, but I guess we can always leave it pull the latest version until we run into problems

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
cburns93: Nothing wrong with that, however just to get this story done we could do it later on? I already have a few things I want to do , lombok was actually one of them, as well as using jetty and speeding up start time

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
jsimons1989: Yes

pom.xml
drosser: I think we should always hard-version our dependencies or avoid running into problems later. Its safer for us

src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
drosser: I think it says that we support a certain type of accounts in this service when in reality we don't

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
drosser: Fine with that, was just a "nice to have" I guess. Maybe we can get a story/spike drafted to look into it and bring it in soon.

pom.xml
cburns93: I disagree only because it forces us to update our libraries as they will never get updated anyways. Additionally if anything does go wrong, we can simply start hard versioning.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public Account createAccount(Account account) {
+        AccountEntity accountEntity = new AccountEntity().build(account);
+
+        AccountEntity createdAccount = accountRepository.insert(accountEntity);
+
+        account = new Account();
+        BeanUtils.copyProperties(createdAccount.getData(), account);
```
jsimons1989: This could be simplified to:
```
// set kind, etags and links on Account
AccountEntity accountEntity = transformAccountEntity (or simply use BeanUtils)
accountRepository.insert(accountEntity);
return account;
```

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
cburns93: It can definitely be added to the list of todos

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
```diff
@@ -0,0 +1,79 @@
...
+        this.id = id;
+        this.etag = GenerateEtagUtil.generateEtag();
+        this.kind = Kind.ACCOUNT.getValue();
+
+        Map<String, String> links = new HashMap<>();
+        
+        links.put(LinkType.SELF.getLink(), "");
+        this.links = links;
+        return this;
+    }
```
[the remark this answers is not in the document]
cburns93: Yeh, that is a discussion me and Steve has had whereas we would use DTOs instead. The reason why the build method exists is because I thought DTOs weren't an option.

pom.xml
drosser: I disagree. If we don't hard-version, and it always pulls the latest, something will inevitably break at some point in the future, and we'll spend a long time figuring out what it was. Versioning is needed, always pulling the latest causes pain points (see: Sarama issues as one example).

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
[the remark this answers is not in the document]
drosser: Lets assume the design can change (because it can) - this seems to be something we should get right now as opposed to creating more work for ourselves later to rework

src/main/java/uk/gov/companieshouse/api/accounts/controller/AccountController.java
```diff
@@ -0,0 +1,31 @@
...
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity createAccount(@Valid @RequestBody Account account) {
+        account = accountService.createAccount(account);
+
+        if (account != null) {
```
drosser: Should we be checking for null here? By explicitly checking for null, aren't we coupling our implementation of a lower level (I think here it's the Service implementation, in which case you're always allocating the object in memory so it would never be null?)

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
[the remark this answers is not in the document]
cburns93: We will raise another story for the DTOs because that in itself needs some investigation.

src/main/java/uk/gov/companieshouse/api/accounts/controller/AccountController.java
cburns93: removed. forgot that was there, it was originally in the service layer for the mongo response, but it only ever responds with the inserted object or a runtime exception.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
cburns93: done

src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
cburns93: removed

pom.xml
cburns93: done

src/main/java/uk/gov/companieshouse/api/accounts/service/AccountService.java
```diff
@@ -0,0 +1,21 @@
...
+ * A service for {@link AccountEntity} and its data {@link AccountDataEntity}
+ */
+public interface AccountService {
+
+    /**
+     * Create the {@link AccountEntity} that will be persisted to the DB
+     * from the given {@link Account} object and returns {@link ResponseEntity}
+     * detailing its success or failure
+     *
+     * @return An {@link ResponseEntity}
```
jsimons1989: Can you update the JavaDoc please

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+
+    public Account createAccount(Account account) {
+        AccountDataEntity accountDataEntity = new AccountDataEntity();
+        AccountEntity accountEntity = new AccountEntity();
+        BeanUtils.copyProperties(account, accountDataEntity);
+
+        //set properties etag, link
+        generateEtagLinksKind(accountDataEntity);
+        accountEntity.setData(accountDataEntity);
```
jsimons1989: Why don't you pass the Account object into generateEtagLinksKind()? That would remove the need for the second BeanUtils.copyProperties()

src/main/java/uk/gov/companieshouse/api/accounts/service/AccountService.java
cburns93: done

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    private AccountRepository accountRepository;
+
+    /**
+     * {@inheritDoc}
+     */
+
+    public Account createAccount(Account account) {
+        AccountDataEntity accountDataEntity = new AccountDataEntity();
+        AccountEntity accountEntity = new AccountEntity();
+        BeanUtils.copyProperties(account, accountDataEntity);
```
macreman: This should be done by calling the versioning/transform service. Under the hood it can use this method but if we leave the entire versioning/transform service out then this POC isn't complete as it doesn't include all aspects of the API implementation for the skeleton balance sheet

cburns93: I have already raised it that we should use DTOs, we are now pending discussion on how we implement that.

cburns93: This has been discussed and will be done under a separate story

cburns93:
Unit tests have been added

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/AccountDataEntity.java
```diff
@@ -0,0 +1,64 @@
...
+import java.time.LocalDate;
+import java.util.HashMap;
+import java.util.Map;
+import javax.validation.constraints.NotNull;
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class AccountDataEntity {
+
+    @Field("id")
+    private String id;
```
macreman: this class is represents an inner 'data' subdocument within a mongodb document so doesn't need an id field, the mongo id is in the wrapper AccountEntity class

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AccountServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+        return account;
+    }
+
+    private void generateEtagLinksKind(Account account) {
+        account.setEtag(GenerateEtagUtil.generateEtag());
+        account.setKind(Kind.ACCOUNT.getValue());
+
+        Map<String, String> links = new HashMap<>();
+
+        links.put(LinkType.SELF.getLink(), "");
```
macreman: I'm assuming this self link isn't set properly due to the interceptor for transaction not being written yet (so we don't have tx_id) so we'll need to rework this later to set an actual value. 
The other part to this is that we'll need to pass in a generated id that we also set on the AccountEntity class during or just after the transform done in the method above.

PR #10, opened 2018-07-19
Add an abstract Service class and an Interface for the Transformers.

sroberts2:
Adds an interface for future Transformer classes to implement.

Adds a base class for extension by future Service classes which contain the common functionality that exists in every Service class implementation. Added the relevant interface to cover this.

Helps resolve: SFA-337

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
@@ -20,8 +20,6 @@ public class CompanyAccountController {
     @PostMapping(value = "/transactions/{transactionId}/company-accounts",
             produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount companyAccount) {
-        companyAccount = companyAccountService.createCompanyAccount(companyAccount);
-
-        return ResponseEntity.status(HttpStatus.CREATED).body(companyAccount);
+        return ResponseEntity.status(HttpStatus.CREATED).body(companyAccountService.save(companyAccount));
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
@@ -1,0 +1,5 @@
+package uk.gov.companieshouse.api.accounts.model.entity;
+
+public interface BaseEntity {
+
+}
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
@@ -1,5 +1,20 @@
 package uk.gov.companieshouse.api.accounts.model.entity;
 
-public interface BaseEntity {
+import org.springframework.data.annotation.Id;
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class BaseEntity {
+
+    @Id
+    @Field("_id")
+    private String id;
+
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
 
 }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
@@ -1,0 +1,42 @@
+package uk.gov.companieshouse.api.accounts.service.impl;
+
+import java.util.Optional;
+import org.springframework.data.mongodb.repository.MongoRepository;
+import org.springframework.stereotype.Service;
+import uk.gov.companieshouse.GenerateEtagUtil;
+import uk.gov.companieshouse.api.accounts.model.entity.BaseEntity;
+import uk.gov.companieshouse.api.accounts.model.rest.RestObject;
+import uk.gov.companieshouse.api.accounts.service.AbstractService;
+import uk.gov.companieshouse.api.accounts.transformer.GenericTransformer;
+
+@Service
+public abstract class AbstractServiceImpl<C extends RestObject, E extends BaseEntity, K> implements
+        AbstractService<C, E, K> {
+
+    private MongoRepository<E, K> mongoRepository;
+
+    private GenericTransformer<E, C> genericTransformer;
+
+    public AbstractServiceImpl(MongoRepository<E, K> mongoRepository,
+            GenericTransformer<E, C> genericTransformer) {
+        this.mongoRepository = mongoRepository;
+        this.genericTransformer = genericTransformer;
+    }
+
+    @Override
+    public C save(C rest) {
+        addEtag(rest);
+        addLinks(rest);
+        addKind(rest);
+        E baseEntity = genericTransformer.transform(rest);
+        mongoRepository.save(baseEntity);
+        return rest;
+    }
+
+    @Override
+    public void addEtag(C rest) {
+        rest.setEtag(GenerateEtagUtil.generateEtag());
+    }
+
+
+}
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
@@ -25,9 +25,10 @@ public abstract class AbstractServiceImpl<C extends RestObject, E extends BaseEn
     @Override
     public C save(C rest) {
         addEtag(rest);
-        addLinks(rest);
         addKind(rest);
+        addLinks(rest);
         E baseEntity = genericTransformer.transform(rest);
+        addID(baseEntity);
         mongoRepository.save(baseEntity);
         return rest;
     }
@@ -37,5 +38,8 @@ public abstract class AbstractServiceImpl<C extends RestObject, E extends BaseEn
         rest.setEtag(GenerateEtagUtil.generateEtag());
     }
 
-
+    @Override
+    public void addID(BaseEntity entity) {
+        entity.setId("generated id");
+    }
 }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
@@ -1,6 +1,5 @@
 package uk.gov.companieshouse.api.accounts.service.impl;
 
-import java.util.Optional;
 import org.springframework.data.mongodb.repository.MongoRepository;
 import org.springframework.stereotype.Service;
 import uk.gov.companieshouse.GenerateEtagUtil;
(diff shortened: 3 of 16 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+        this.genericTransformer = genericTransformer;
+    }
+
+    @Override
+    public C save(C rest) {
+        addEtag(rest);
+        addLinks(rest);
+        addKind(rest);
+        E baseEntity = genericTransformer.transform(rest);
+        mongoRepository.save(baseEntity);
```
cburns93: Would this work if we where to add the setting of the ID? currently we allow MongoDB to set the IDs but as we are moving away from that, we will need to set them just after the transform from the rest and just before the save to the database

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -20,8 +20,6 @@
     @PostMapping(value = "/transactions/{transactionId}/company-accounts",
             produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount companyAccount) {
-        companyAccount = companyAccountService.createCompanyAccount(companyAccount);
-
-        return ResponseEntity.status(HttpStatus.CREATED).body(companyAccount);
+        return ResponseEntity.status(HttpStatus.CREATED).body(companyAccountService.save(companyAccount));
```
cburns93: If we are calling the generic `save` where would the validation fit in with this solution?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+    private GenericTransformer<C, E> genericTransformer;
+
+    public AbstractServiceImpl(MongoRepository mongoRepository,
+            GenericTransformer<C, E> genericTransformer) {
+        this.mongoRepository = mongoRepository;
+        this.genericTransformer = genericTransformer;
+    }
+
+    @Override
+    public C save(C rest) {
```
cburns93: What does `C` stand for? It doesn't seem to follow the standard generic type parameter conventions?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
sroberts2: Yes. An addID method would be added to the AbstractService class - then depending on if the functionality is common it would be implemented in the AbstractServiceImpl class, if it's not common functionality it would be implemented in the relevant service implementation class.

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.companieshouse.api.accounts.model.entity;
+
+public interface BaseEntity {
+
```
cburns93: I believe we will have to add `ID` into this interface?

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
sroberts2: Using a generic save will not effect the functionality of the validation implementation.

cburns93: Where would the validation fit in?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
sroberts2: Suggest different letters and I will change them

cburns93: It's not that you should change them, I'm just trying to gauge what C represents?

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
sroberts2: I wasn't sure that the ID implementation was set in stone, happy to add it if it is.

cburns93: Yeah, its set in stone,  we just need to implement it now.

cburns93:
Detailed JavaDoc would be a good addition also.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -2,47 +2,36 @@
...
     }
 
-    private void generateEtagLinksKind(CompanyAccount companyAccount) {
-        companyAccount.setEtag(GenerateEtagUtil.generateEtag());
-        companyAccount.setKind(Kind.ACCOUNT.getValue());
-
+    @Override
+    public void addLinks(CompanyAccount rest) {
         Map<String, String> links = new HashMap<>();
+        links.put(LinkType.SELF.getLink(), "self link");
```
cburns93: How would you set the link on the REST object if it does not contain the ID yet? (populated after the REST to entity transformation - before the save to mongo).

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
sroberts2: This conversation should not be conducted via the PR process as it involves specific functionality of another piece of work, but if I was to take a stab at it without any investigation I would add an interface along the lines of 'public interface validation<T>'. Then add it into the AbstractServiceImpl class performing constructor injection to get a specific implementation in the same manner that we use for the GenericTransformer.

cburns93: I'm just making sure we don't commit to any structure that makes it harder for ourselves in future. If you believe it won't be a problem, then that's all I need to know.

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/BaseEntity.java
sroberts2: Done.

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
jsimons1989: I know it's part of another piece of work but this wouldn't allow returning error HTTP responses if the validation fails or an error occurs inside companyAccountService.save when you have already set the return status code before calling save. One option would be to validate inputs before calling the service to handle the creation which would work with your existing code. I am assuming the global error handler catches all errors and returns the appropriate error code?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -2,47 +2,36 @@
...
 import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
 import uk.gov.companieshouse.api.accounts.model.rest.CompanyAccount;
-import uk.gov.companieshouse.api.accounts.repository.CompanyAccountRepository;
 import uk.gov.companieshouse.api.accounts.service.CompanyAccountService;
-import uk.gov.companieshouse.api.accounts.transformer.CompanyAccountTransformer;
+import uk.gov.companieshouse.api.accounts.transformer.GenericTransformer;
 
 @Service
-public class CompanyAccountServiceImpl implements CompanyAccountService {
+public class CompanyAccountServiceImpl extends AbstractServiceImpl<CompanyAccount, CompanyAccountEntity> implements CompanyAccountService  {
```
jsimons1989: CompanyAccountServiceImpl implementing CompanyAccountService and extending a AbstractServiceImpl which implements a AbstractService feels a bit overkill. Why are so many interfaces/abstract classes needed?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
jsimons1989: The self link can be added when needed not populated with made up data. Otherwise, It's more likely to get missed later.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
sroberts2: I wanted to add some generic method interfaces that every Service should implement such as addLinks, addEtag, addKind etc so I created the AbstractService that every service should extend from. I then realised that for some methods like addEtag every implementation would be the same so I added the AbstractServiceImpl to provide them.

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
sroberts2: This will not work for validation as is, the only error handling that is currently implemented correctly is the Bind error for the CompanyAccount class. All other errors would be in another PR we currently only return status 201 created. The purpose of this PR is to get the AbstractService in place so I can incrementally build on top of it.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
jsimons1989: Can't you use one abstract class or an interface with default methods if every service must implement addLinks, etc?

What is the CompanyAccountService interface adding? Why can't you simplify to:
```
public class CompanyAccountService extends AbstractService<CompanyAccount, CompanyAccountEntity>{
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -2,47 +2,36 @@
...
-    private void generateEtagLinksKind(CompanyAccount companyAccount) {
-        companyAccount.setEtag(GenerateEtagUtil.generateEtag());
-        companyAccount.setKind(Kind.ACCOUNT.getValue());
-
-        Map<String, String> links = new HashMap<>();
+    @Override
+    public void addLinks(CompanyAccount rest) {
+//        Map<String, String> links = new HashMap<>();
+//        links.put(LinkType.SELF.getLink(), "");
+//        rest.setLinks(links);
```
jsimons1989: Can you remove the commented out code or add a TODO linked to a story related to it please?

Commented out code is bad practise because it confuses devs in the future whether it was meant to be added, removed or doesn't work. See: https://companieshouse.atlassian.net/wiki/spaces/DEV/pages/142213166/Readable+Maintainable+Testable+Elegant+Code#Readable,Maintainable,Testable,ElegantCode-Comments

PR #11, opened 2018-07-23
WIP - API logging

gpatrac:
SFA-463 - API Logging using Spring AOP. The controller aspect logs entry and exit into the controller method along with the response time at exit.


src/main/java/uk/gov/companieshouse/api/accounts/utility/logging/AccountsLoggingAspect.java
```diff
@@ -0,0 +1,84 @@
...
+          String.format("Exited : %s with return value %s", getMethodName(joinPoint), retVal));
+    }
+  }
+
+  @Around("controllerMethods() || serviceMethods()")
+  public void logControllerPerformanceStats(ProceedingJoinPoint joinPoint) throws Throwable {
+    long startTime = System.currentTimeMillis();
+
+    //Do not change, the target method will not progress without this.
+    joinPoint.proceed();
```
jterrych: this advice will swallow all returning objects and it looks like it will run for every controller/service method. If its intended to run for all methods then joinPoint.proceed() should be returned. And if thats the case then having @Before and @After will add unneeded advices as the @Around covers it.

src/main/java/uk/gov/companieshouse/api/accounts/utility/logging/AccountsLoggingAspect.java
```diff
@@ -0,0 +1,84 @@
...
+
+    if (apiLogging.isPerformanceStatsEnabled()) {
+      LOG.debug(String.format("The method %s Started At %s",
+          getMethodName(joinPoint), startTime));
+      LOG.debug(String.format("%s Completed in: %s milliseconds",
+          getMethodName(joinPoint), timeTaken));
+    }
+  }
+
+  @AfterThrowing(pointcut = "controllerMethods() || serviceMethods()", throwing = "exception")
```
jterrych: seems a little weird to have the around advice throw the exception that will again be advised here, unless more fine tuning of pointcuts are planned?

src/main/java/uk/gov/companieshouse/api/accounts/utility/logging/AccountsLoggingAspect.java
gpatrac: Depends on the decision on what should we be able to log and when. The @around is logging performance stats only. We may need to log from service methods while the before and after is for method trace only and could include the logging of the request and response to help debug issues.

pom.xml
```diff
@@ -29,8 +29,9 @@
         <api-helper-java.version>1.0.0-rc1</api-helper-java.version>
         <structured-logging.version>1.2.0-rc1</structured-logging.version>
         <aws-java-sdk.version>1.11.52</aws-java-sdk.version>
-        <environment-reader-library.version>1.0.0-rc1</environment-reader-library.version>
+        <environment-reader-library.version>1.2.0-rc1</environment-reader-library.version>
         <jackson-datatype-jsr310.version>2.9.6</jackson-datatype-jsr310.version>
+        <aspcetj-version>1.8.0</aspcetj-version>
```
cburns93: spelling of aspectj. additionally can we prefix version with full stop like we've done elsewhere

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -17,12 +17,12 @@
 
 @SpringBootApplication
 public class CompanyAccountsApplication implements WebMvcConfigurer {
+    public static final String APPLICATION_NAME_SPACE = "company-accounts.api.ch.gov.uk";
```
cburns93: Good idea, I'm sick off seeing the same string duplicated in each instance of where we defined a logger. :+1:

src/test/java/uk/gov/companieshouse/api/accounts/utility/ApiLoggingTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.jupiter.api.Test;
+import uk.gov.companieshouse.api.accounts.utility.logging.ApiLogging;
+
+public class ApiLoggingTest {
+
+  /**
+   * Test that the apiLogging.methodTraceEnabled defaults to False
+   */
+  @Test
+  public void methodTraceNotEnabled() {
```
cburns93: Since we are using JUnit5, could we put `@DisplayName` on each test just to give a descriptive name of what it does. This then can replace the JavaDoc description. :)

src/test/java/uk/gov/companieshouse/api/accounts/utility/TestEnvironmentSetupHelper.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.companieshouse.api.accounts.utility;
+
+import java.lang.reflect.Field;
+import java.util.Map;
+
+public interface TestEnvironmentSetupHelper {
```
cburns93: Perhaps put a JavaDoc description somewhere in this class to detail that it can be used for testing to set environment variables?

cburns93:
Could we have a more a descriptive description on what this PR contains?

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
```diff
@@ -0,0 +1,87 @@
...
+    LogBuilder logBuilder = getLogBuilder(joinPoint);
+    logBuilder.logStartOfRequestProcessing(String.format("Start Request %s",
+        methodName(joinPoint)));
+
+    logBuilder.logEndOfRequestProcessing(String.format("End of Request %s Successfully Completed",
+        methodName(joinPoint)), statusCode(result), responseTime);
+  }
+
+  private String methodName(JoinPoint joinPoint) {
+    return joinPoint.getSignature().getName();
```
jsimons1989: Some of the controller methods are called "create" in multiple controllers. How will this be easy to debug?

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/LogBuilderImpl.java
```diff
@@ -0,0 +1,65 @@
...
+    this.transactionId = transactionId;
+    this.accountsId = accountsId;
+  }
+
+  public void logStartOfRequestProcessing(String message) {
+    Map<String, Object> logData = new HashMap();
+    addValues(logData);
+    putValue(logData, "message", message);
+    putValue(logData, "identity", userId);
+    LOGGER.infoContext(requestId, "", logData);
```
jsimons1989: A builder shouldn't be sending the log messages. A build is responsible for building complex objects.

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Helper to get the HttpServletRequest so that we can get the requestID and the userID.
+   */
+  private HttpServletRequest getRequest(JoinPoint joinPoint) {
+    return (HttpServletRequest) joinPoint.getArgs()[0];
+  }
+
+  private String transactionId(JoinPoint joinPoint) {
+    return joinPoint.getArgs()[1].toString();
```
jsimons1989: Is this forcing controller method signatures to be a certain format?
```java
public void [name](request, transactionId, ...)
```

The AOP shouldn't be controlling the method signatures. Will this work for other controllers? e.g.
```java
public ResponseEntity create(@Valid @RequestBody SmallFull smallFull, HttpServletRequest request)
```

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private int statusCode(Object result) {
+    assert(result instanceof ResponseEntity);
+    return ((ResponseEntity) result).getStatusCode().value();
```
jsimons1989: What happens when an AssertionError is thrown? The globalExceptionHandler is catching Exceptions but not Errors. Will this be correctly handled and clearly logged?

From [Oracle docs](https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html#usage):

> Another problem with using assertions for argument checking is that erroneous arguments should result in an appropriate runtime exception (such as IllegalArgumentException, IndexOutOfBoundsException, or NullPointerException). An assertion failure will not throw an appropriate exception.

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
gpatrac: agreed, the request parameters can be identified based on their type, ex HttpServletRequest, SmallFull etc. However for the Ids (transactionId, accountId) should we log them as an array of Strings associated with a key=resources ?

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
gpatrac: will change it

gpatrac:
Tests added for AccountsLogger and AOP

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
```diff
@@ -0,0 +1,108 @@
...
+    logger.logStartOfRequestProcessing(String.format(AccountsUtility.START_OF_RQUEST_MSG.value(),
+        methodName));
+
+    long startTime = System.currentTimeMillis();
+    //Do not change, the target method will not progress without this.
+    Object result = null;
+
+    try {
+      result = joinPoint.proceed();
+    } catch (Throwable t) {
```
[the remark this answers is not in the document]
gpatrac: The exceptions are handled by GlobalExceptionHandler. The only Exceptions that are not handled by GlobalExcetionHandler would ever be logged here.

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
```diff
@@ -0,0 +1,108 @@
...
+  private AccountsLogger getLogger(JoinPoint joinPoint) {
+    HttpServletRequest request = getRequest(joinPoint);
+    return new AccountsLoggerImpl(requestId(request), userId(request), pathVariables(joinPoint));
+  }
+
+  /**
+   * Helper to get the HttpServletRequest so that we can get the requestID and the userID.
+   */
+  private HttpServletRequest getRequest(JoinPoint joinPoint) {
+    return (HttpServletRequest) Arrays.stream(joinPoint.getArgs())
```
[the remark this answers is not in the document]
gpatrac: done

src/main/java/uk/gov/companieshouse/api/accounts/util/logging/ControllerLoggingAspect.java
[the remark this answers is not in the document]
jterrych: ok. +1

PR #13, opened 2018-07-24
Adding small full IXBRL model

ch-mcabrera:
Changes to add the basic IXBRL model that will be used by the filing generator to create the JSON which will be sent **document render service.**

This initial commit contains one field from the balance sheet and one note; and it will create the foundation for adding the other balance sheet's fiends and notes in the future. 

_**Relevant Links**_
[Filing Generator Design]( https://companieshouse.atlassian.net/wiki/x/gYAtGg)
[XBRL Model](https://companieshouse.atlassian.net/wiki/x/aQBjGg)

Resolves:     SFA-122

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
@@ -20,8 +20,6 @@ public class CompanyAccountController {
     @PostMapping(value = "/transactions/{transactionId}/company-accounts",
             produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount companyAccount) {
-        companyAccount = companyAccountService.createCompanyAccount(companyAccount);
-
-        return ResponseEntity.status(HttpStatus.CREATED).body(companyAccount);
+        return ResponseEntity.status(HttpStatus.CREATED).body(companyAccountService.save(companyAccount));
     }
 }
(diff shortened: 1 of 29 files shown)


ch-mcabrera:
Regarding your comment: 
```
Can we sort the model into an iXBRL package within model? smallfull will confuse readers what it contains as technically there are also small full models within the entity and rest packages.
```
I didnt want to use the IXBRL name in the package so I decided to use same naming convention we currenlty use in Abridge for the model. 
And in the future, we might have `small-full` and `Abridged` model within model's package.

Abridge Model: 
https://github.com/companieshouse/abridged.accounts.api.ch.gov.uk/tree/develop/src/main/java/uk/gov/companieshouse/api/accounts/model/abridged

ch-mcabrera:
@cburns93

Regarding 
> Additionally, talk to @BSaunders95 about using Project Lombok to get rid of the getters or setters

Has this been decided to go down this route, to use Project Lombok?

BSaunders95:
We're going to have a discussion with Leads and architects next week to decide whether to use Lombok.

ch-mcabrera:
@cburns93 
>I see that the model package has changed to iXBRL, you mentioned that there would be small-full and abridged models inside, how would you distinguish between the two?

This is something that hasn't been discussed yet. 
Besides, we are in the same situation with the ixbrl as the other packagers under `model`: _`entity`_ and _`rest`_. They are all for `small-full` and, as i said, this package naming might change once we introduce `abridged` and it will impact all these three packages: entity, rest and ixbrl

PR #16, opened 2018-07-25
Added id generation for the database

sroberts2:
Added the ID generation for existing entities.
Added a global error handling class as a result due to a NoSuchAlgorithmException needing to be captured.

The account number which should be read from the Transaction is currently unavailable so a TODO is added in both existing Controller classes.

Helps resolve: SFA-337

PR #18, opened 2018-07-26
Connecting accounts to transaction resource links

cburns93:
- adds new company-account resource to the transaction resource links
- adds new functionality to deal with the patching of the transaction using api key
- amended company accounts service layer implementation to deal with the patching of the transaction
- minor refactor of other classes and interfaces (includes removing redundant code)
- adds and amends unit tests to cover new and amended functionality

**NOTE:** The transaction patching implementation will be taken out and put into the private sdk, so it is worth noting that the currently implementation is only temporary.

Resolves: SFA-465


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,33 +1,98 @@
...
+
+    public void addEntityLinks(CompanyAccountEntity entity, ResponseEntity<Transaction> transaction) {
+        Map<String, String> map = new HashMap<>();
+        String selfLink = createCompanyAccountSelfLink(entity, transaction);
+        map.put(LinkType.SELF.getLink(), selfLink);
+        entity.getData().setLinks(map);
+    }
+
+    private String createCompanyAccountSelfLink(CompanyAccountEntity entity, ResponseEntity<Transaction> transaction) {
+        return getTransactionSelfLink(transaction) + "/company-accounts/" + entity.getId() + "";
```
jsimons1989: An empty string is being added to the link?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,33 +1,98 @@
...
+        addEtag(companyAccount);
+
+        CompanyAccountEntity companyAccountEntity = genericTransformer.transform(companyAccount);
+
+        addEntityID(companyAccountEntity);
+        addEntityLinks(companyAccountEntity, transaction);
+
+        mongoRepository.insert(companyAccountEntity);
+
+        transactionManager.updateTransaction(transaction.getBody().getId(), request.getHeader("X-Request-Id"), getCompanyAccountSelfLink(companyAccountEntity));
```
jsimons1989: This could be simplified to:
```java
String companyAccountsLink = getCompanyAccountsLink(transaction, entity.getId());
addLinks(companyAccountEntity);
mongoRepository.insert(companyAccountEntity);
transactionManager.updateTransaction(transaction.getBody().getId(), request.getHeader("X-Request-Id"), companyAccountsLink);
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,33 +1,98 @@
...
+    public void addEntityID(CompanyAccountEntity entity) {
+        entity.setId(UUID.randomUUID().toString());
+    }
+
+    private static HttpServletRequest getRequestFromContext() {
+        return ((ServletRequestAttributes)RequestContextHolder.getRequestAttributes()).getRequest();
+    }
+
+    private static ResponseEntity<Transaction> getTransactionFromSession(HttpServletRequest request) {
+        return (ResponseEntity) request.getSession().getAttribute("transaction");
```
jsimons1989: "transaction" is being hard coded in multiple places in this PR and https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/16. It would be better to place it in a separate class to avoid issues in the future if someone changes it.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
```diff
@@ -34,6 +35,10 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
             String transactionId = pathVariables.get("transactionId");
             ResponseEntity<Transaction> transaction = transactionManager
                     .getTransaction(transactionId, request.getHeader("X-Request-Id"));
+
+            HttpSession session = request.getSession();
+            session.setAttribute("transaction", transaction);
```
jsimons1989: This PR places the transaction in session scope but https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/16 is placing the transaction in the request scope. Can you agree a consistent approach please?

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
```diff
@@ -1,42 +1,176 @@
...
 
-    private RestTemplate transactionRestTemplate;
-
     @Autowired
     private TransactionServiceProperties configuration;
 
     private static final String ID_PARAMETER = "{id}";
+    private static final String X_REQUEST_ID = "X-Request-Id";
+
+    // represents the Authorization header name in the request
```
jsimons1989: Can you make this into a JavaDoc rather than a comment please?
```java
/* represents the Authorization header name in the request */
private static final String AUTHORIZATION_HEADER = "Authorization";
```

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
```diff
@@ -1,42 +1,176 @@
...
+    }
+
+    /**
+     * Patch the transaction resource
+     *
+     * @param url - url to send the get request
+     * @param requestEntity - the request entity object
+     */
+    private void patchTransaction(String url, HttpEntity requestEntity) {
+        restTemplate.exchange(getRootUri() + url, HttpMethod.PATCH, requestEntity, Void.class);
```
jsimons1989: You are not checking if the patch was successful.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
cburns93: Yeh good spot. Fixed in [84aa56c](https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/18/commits/84aa56c9f1d5acd2ce6b9a79332a164d195a364d)

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
cburns93: I can't use the overriden addLinks() method because of the way the abstract service is set up. I will have to change the abstract service, which will break the small full stuff. Additionally, the entityLinks has to be after the setting of the id of the company-accounts resource

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
cburns93: Will be fixed by Steve in a different PR

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
cburns93: Is this still an issue after what Steve will do in his PR?

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
cburns93: How would we check if the patch was succesfull?

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
jsimons1989: Nope, Steve is going to modify his PR to be consistent with your approach.

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManager.java
```diff
@@ -5,4 +5,6 @@
 public interface TransactionManager {
```
sroberts2: I was under the impression all the classes in the transaction package were copied form the Transaction API and were to be deleted at some stage. If we start changing them don't we run the risk of adding functionality that prevents us from doing this?

cburns93: Which parts are you referencing has been changed?

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
cburns93: What would be in the `ResponseEntity` type? 
`ReponseEntity<Transaction> resp = restTemplate.exchange(getRootUri() + url, HttpMethod.PATCH, requestEntity, Void.class);`

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
sroberts2: The addLinks() is my fault I wanted to get some early visibility of the sort of methods we would need to implement, I have a branch of code where I revisit that due for PR imminently. I do agree with Jo's simplification of the code though that makes the last call to genericTransformer.transform(companyAccountEntity) unnecessary.

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
jsimons1989: It would be ResponseEntity<Void> but I believe you can use ResponseEntity<?> as you don't care about the type.

cburns93: I've inserted the logic for the checking of the status code, but the actual error to be throw will be covered by the api error story

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
cburns93: done

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CurrentPeriodServiceImpl.java
```diff
@@ -32,6 +32,11 @@ public void addKind(CurrentPeriod rest) {
         rest.setKind(Kind.CURRENT_PERIOD.getValue());
     }
 
+    @Override
+    public void addID(CurrentPeriodEntity entity) {
+
```
jsimons1989: We shouldn't be overriding with empty methods. Are these going to have any functionality?

cburns93: fixed

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,46 +1,98 @@
...
+        addLinks(companyAccount, companyAccountLink);
+
+        CompanyAccountEntity companyAccountEntity = companyAccountTransformer.transform(companyAccount);
+
+        companyAccountEntity.setId(id);
+        companyAccountRepository.insert(companyAccountEntity);
+
+        transactionManager.updateTransaction(transaction.getId(),
+                request.getHeader("X-Request-Id"),
+                getCompanyAccountSelfLink(companyAccountEntity));
```
jsimons1989: The companyAccountLink can be used here instead of needing to implement a new method to get the self link

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImplTest.java
```diff
@@ -2,54 +2,116 @@
...
+        doReturn(companyAccountEntityMock).when(companyAccountTransformer).transform(ArgumentMatchers
+                .any(CompanyAccount.class));
+
+        when(companyAccountEntityMock.getData()).thenReturn(companyAccountDataEntityMock);
+
+        CompanyAccount result = companyAccountService.createCompanyAccount(companyAccountMock);
+  
         assertNotNull(result);
-        assertEquals(companyAccount, result);
+        assertEquals(companyAccountMock, result);
```
jsimons1989: This test is titled that it tests the successful creation of the resource but it doesn't verify that the resource was created.

e.g.
```java
// set up mocks
// run test

assertNotNull(result);
assertEquals(companyAccountMock, result);
verify(companyAccountRepository.save(companyAccountEntityMock));
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CurrentPeriodServiceImpl.java
cburns93: I'm not sure why these changes are coming up as I have not edited the files at all, however,I have removed the method

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImplTest.java
```diff
@@ -2,54 +2,116 @@
...
+     * creates an open or closed dummy transaction depending on the boolean passed into method
+     *
+     * @param isOpen - true = open, false - closed
+     * @return ResponseEntity<> with the desired transaction
+     */
+    private Transaction createDummyTransaction(boolean isOpen) {
+        Transaction transaction = new Transaction();
+        transaction.setId("id");
+        transaction.setStatus(isOpen ? TransactionStatus.OPEN.getStatus() : TransactionStatus.CLOSED.getStatus());
+        transaction.setLinks(createLinksMap());
```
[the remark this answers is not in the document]
cburns93: done

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImplTest.java
cburns93: done

src/main/java/uk/gov/companieshouse/api/accounts/transaction/TransactionManagerImpl.java
```diff
@@ -1,42 +1,180 @@
...
+    /**
+     * Patch the transaction resource
+     *
+     * @param url - url to send the get request
+     * @param requestEntity - the request entity object
+     */
+    private void patchTransaction(String url, HttpEntity requestEntity) {
+        ResponseEntity<?> response = restTemplate.exchange(getRootUri() + url, HttpMethod.PATCH, requestEntity, Void.class);
+        if (response.getStatusCode() != HttpStatus.OK) {
+            // insert error exceptions here
```
jsimons1989: This error needs to be handled. This could be missed in the future.

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImplTest.java
```diff
@@ -2,54 +2,112 @@
...
+        CompanyAccount result = companyAccountService.createCompanyAccount(companyAccountMock);
+  
         assertNotNull(result);
-        assertEquals(companyAccount, result);
+        assertEquals(companyAccountMock, result);
+        verify(companyAccountRepository).insert(companyAccountEntityMock);
+    }
+
+    /**
+     * creates an open or closed dummy transaction depending on the boolean passed into method
```
jsimons1989: This JavaDoc does not match the method. Can you update the JavaDoc please?

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -24,11 +21,8 @@
...
             produces = MediaType.APPLICATION_JSON_VALUE)
     public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount companyAccount,
-            HttpServletRequest request)
-            throws NoSuchAlgorithmException {
-        Transaction transaction = (Transaction) request.getSession().getAttribute(AttributeName.TRANSACTION.getValue());
-        CompanyAccount result = companyAccountService
-                .save(companyAccount, transaction.getCompanyNumber());
+            HttpServletRequest request) {
+        CompanyAccount result = companyAccountService.createCompanyAccount(companyAccount);
         return ResponseEntity.status(HttpStatus.CREATED).body(result);
```
macreman: we cannot hard code a successful response code, we have to validate the creation of the company accounts resource

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,46 +1,93 @@
...
+    @Autowired
+    private CompanyAccountRepository companyAccountRepository;
+
+    @Autowired
+    private CompanyAccountTransformer companyAccountTransformer;
+
+    /**
+     * {@inheritDoc}
+     */
+    public CompanyAccount createCompanyAccount(CompanyAccount companyAccount) {
```
macreman: we should be passing the request into this service so it is available to get the transaction and for use in logging for following requests through the service

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,46 +1,93 @@
...
+    private String createSelfLink(Transaction transaction, String id) {
+        return getTransactionSelfLink(transaction) + "/company-accounts/" + id;
+    }
+
+    private String getTransactionSelfLink(Transaction transaction) {
+        return transaction.getLinks().get(LinkType.SELF.getLink());
+    }
+    
+    private void addKind(CompanyAccount rest) {
+        rest.setKind("company-accounts");
```
macreman: this kind should be `company-accounts#company-accounts` as this is what we've decided for the `small-full-accounts#small-full-accounts`

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,46 +1,93 @@
...
+        String id = generateID();
+        String companyAccountLink = createSelfLink(transaction, id);
+        addKind(companyAccount);
+        addEtag(companyAccount);
+        addLinks(companyAccount, companyAccountLink);
+
+        CompanyAccountEntity companyAccountEntity = companyAccountTransformer.transform(companyAccount);
+
+        companyAccountEntity.setId(id);
+        companyAccountRepository.insert(companyAccountEntity);
```
macreman: we need to have some error handling on this - what if the database is not there or misconfigured? or if a document already exists with this id?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
sroberts2: I think it would be bad practice to pass the request into the service layer, we should maintain a separation of concerns by extracting what we need and passing that instead.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -1,46 +1,93 @@
...
+
+        transactionManager.updateTransaction(transaction.getId(),
+                request.getHeader("X-Request-Id"),
+                companyAccountLink);
+
+        return companyAccount;
```
macreman: we should not be returning the 'rest' model that we passed into this method, we need the ability to return multiple things here i.e. the response code and body.

Brain dump from several of us:

The easiest way to do this is to create the ResponseEntity with the correct status code and body within the service and return that - this means any errors from the db insert or transaction patch can explicitly set their response codes rather than hoping the global exception handler gets it right.
As the Controller should be dealing with the response codes and body to return we could create a utility object or something similar that can have multiple fields that get mapped to the response code/body by the controller.

Another way to do this would be to throw different exceptions for each case and catch those in the controller which will the decide the response code and body to return if necessary however that is not very nice or explicit and require more logic in the controller.

sroberts2: I like the second method more again best practice in attempting to maintain a clear separation of concerns. You have to think that each Service may be used by multiple Controllers and attempting to code for each Controllers needs in the service layer will add to much complexity.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
macreman: That depends on which was we go with the return value and exception handling from service back to controller (comment above).

If we use typed exceptions to determine response codes in the controller then the service knows nothing about the request so thats fine - the only issue then is that the structured logging will not have access to the request for tracing it through the service.
Personally I'd prefer useful logs that allow us to trace requests through the controller and service.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
macreman: I agree, just means that the controllers will have more logic in them - which is fine.
If this is the way we go then the controller needs to explicitly handle each distinct sunny day and failure case e.g. in this specific function it should handle :
- successful write of company-accounts data and successful patch of transaction
- failed write of company-accounts to mongo
- failed transaction patch
Each of these should be explicitly caught/checked in the controller, maybe requiring this service method to be split up so the controller makes multiple calls into the service,possibly 1 for decorating the REST object with kind, links, etag, another for transformation to DB model and writing to mongo and another for patching transaction

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
sroberts2: We shouldn't be using the Structured logging lib at all in place of an established industry standard logging framework, but that aside if Structured logging needs something in the service layer from the request it should be extracted and passed in.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
sroberts2: The Controllers will have to interpret anything that is returned from the service layer or below as is their job, it follows the throw early, catch late paradigm by design. I agree with outlined functionality detailed above but the error handling isn't part of this PR, we need to address that when we have more of the system in place this is just to complete the logic to get a happy path in place for future work to build on.

macreman: Error handling is part of every PR and should not have to be back ported into code that is already in develop. 
It should be included here even if it's a basic try-catch with a generic exception that will be replaced with more granular exception handling per case later.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
macreman: We have to output structured JSON logs from our microservices and be able to trace requests through all of our services using the request ID from headers.

If there is an industry standard way of doing this without using our structured-logging library (I'm sure there will actually be many ways of doing this) then we can look at using it instead of our structured-logging library.
Maybe worth a look into options going forwards but, as of right now, the structured-logging library is the library that all of our java microservices use, including this one.

As a minimum to get this functionality with the current set up we need to pass in the requestID (rather than the entire request object) to the service so that it can use it for traceability through the logs.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
cburns93: I will throw generic errors at the time being, and then we can improve those error types in the future. At least there's something there to cover the scenarios.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
sroberts2: The Structured logging lib is effectively just System.out.println and there is a reason the industry considers it unsafe or poor practice to introduce System.out into production code. I know the volumes of traffic we are expecting is low in general terms and that it provides the benefit of consistency in logging but we should still be very aware of the fundamental issues we are introducing referencing System in a production environment.

macreman:
fix the conflict in pom.xml and this looks good to go 👍

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
macreman: Agreed, logging to System.out isn't ideal and we should look at improving our structured logging functionality to avoid doing that whether that means changing the underlying implementation in our structured-logging library or moving to something else that gives us the same functionality. 
As I said previously if there is a better industry standard that gives us the same functionality then we can look at switching, if you have any proposals for this we can go through it offline and bring the relevant people into the conversation.

PR #23, opened 2018-07-30
Add the remaining business logic interceptors for the initial build.

sroberts2:
Added the findByExample method to the AbstractService and its Implementation.
Added the SmallFullInterceptor.
Added the CompanyAccountInterceptor.

helps resolve: SFA-509


src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,38 @@
...
+@Component
+public class CompanyAccountInterceptor extends HandlerInterceptorAdapter {
+
+    @Autowired
+    private CompanyAccountService companyAccountService;
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) {
```
jsimons1989: This doesn't match the design. An interceptor is designed to block invalid requests. It should:
1. Get the {companyAccountId} from the request parameters
2. Retrieve the company account from the DB by the ID using the CompanyAccountService
3. Get the stored transaction from the session and verify the company account self link exists in the transaction links
4. Store the company account in the session

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+import uk.gov.companieshouse.api.accounts.AttributeName;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountDataEntity;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
+import uk.gov.companieshouse.api.accounts.service.CompanyAccountService;
+
+@Component
+public class CompanyAccountInterceptor extends HandlerInterceptorAdapter {
```
jsimons1989: Can JavaDocs be added to the interceptors to explain what they do please. There are so many of them in this API and the transactions API that it could be confusing to developers what each one is doing.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
```diff
@@ -37,6 +39,15 @@ public C save(C rest, String companyAccountId) throws NoSuchAlgorithmException {
         return rest;
     }
 
+    @Override
+    public E findByExample(E example) {
+        Optional<E> optional = (Optional<E>) mongoRepository.findOne(Example.of(example));
+        if (optional.isPresent()) {
+            return optional.get();
+        }
+        return null;
```
jsimons1989: Records should be retrieved out of the database using the ID. For child elements (e.g. small-full) generate the expected ID using the transaction and account ID.

src/test/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptorTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @BeforeEach
+    public void setUp() {
+        when(httpServletRequest.getSession()).thenReturn(session);
+        when(httpServletRequest.getRequestURI()).thenReturn("/transactions/123/company-accounts/456");
+        when(companyAccountService.findByExample(any(CompanyAccountEntity.class))).thenReturn(companyAccountEntity);
+    }
+
+    @Test
+    @DisplayName("Tests the interceptor returns correctly after a specific URI is provided")
+    public void testReturnsTheCorrectEntityForTheURI() {
```
jsimons1989: The unit tests need to test the failure scenarios too. The boolean returned from the interceptor should be checked by an assertion.

src/test/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptorTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.mockito.quality.Strictness;
+import org.springframework.web.servlet.HandlerMapping;
+import uk.gov.companieshouse.api.accounts.AttributeName;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountDataEntity;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
+import uk.gov.companieshouse.api.accounts.service.CompanyAccountService;
+import uk.gov.companieshouse.api.accounts.transaction.Transaction;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
cburns93: Can we avoid reducing the strictness of Mockito. We could be suppressing valid warnings if we use Lenient. Mockito themselves say "Strictness drives cleaner tests and better productivity".

cburns93: I think if we remove this line, it will go to the default strictness which is `Strict`. _"STRICT_STUBS - ensures clean tests, reduces test code duplication, improves debuggability. Best combination of flexibility and productivity. Highly recommended. Planned as default for Mockito v3."_

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+                .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
+        if (transaction != null && companyAccountEntity != null) {
+            String companyNumber = transaction.getCompanyNumber();
+            String smallFullId = smallFullService.generateID(companyNumber);
+            SmallFullEntity smallFull = smallFullService.findById(smallFullId);
+
+            if (smallFull != null) {
+                String companyAccountLink = companyAccountEntity.getData().getLinks()
+                        .get("small_full_accounts");
+                String smallFullSelf = smallFull.getData().getLinks().get("self");
```
cburns93: I believe we already have a `self` link in the LinkType : `LinkType.SELF.getLink()`

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,58 @@
...
+                .getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+        String companyAccountId = pathVariables.get("companyAccountId");
+        CompanyAccountEntity companyAccountEntity = companyAccountService
+                .findById(companyAccountId);
+        Transaction transaction = (Transaction) request.getSession()
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        HttpSession session = request.getSession();
+        if (transaction != null && companyAccountEntity != null) {
+            String accountsLink = transaction.getLinks().get("company_account");
+            String accountsSelf = companyAccountEntity.getData().getLinks().get("self");
```
cburns93: I believe we already have a `self` link in the LinkType : `LinkType.SELF.getLink()`

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
sroberts2: Changed

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+        Map<String, String> pathVariables = (Map) request
+                .getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+        String companyAccountId = pathVariables.get("companyAccountId");
+        CompanyAccountEntity companyAccountEntity = companyAccountService
+                .findById(companyAccountId);
+        Transaction transaction = (Transaction) request.getSession()
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        HttpSession session = request.getSession();
+        if (transaction != null && companyAccountEntity != null) {
+            String accountsLink = transaction.getLinks().get("company_account");
```
jsimons1989: Can "company_account" be added to the LinkType?

sroberts2: Yes, the LinkType changes have been merged to develop so I will use that value.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,59 @@
...
+            String accountsSelf = companyAccountEntity.getData().getLinks().get(LinkType.SELF.getLink());
+            if (accountsLink.equals(accountsSelf)) {
+                session.setAttribute(AttributeName.COMPANY_ACCOUNT.getValue(),
+                        companyAccountEntity);
+                return true;
+            }
+        }
+        return false;
```
jsimons1989: Logging is needed here to help debug failed requests

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) throws NoSuchAlgorithmException {
+        HttpSession session = request.getSession();
+        Transaction transaction = (Transaction) request.getSession()
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        CompanyAccountEntity companyAccountEntity = (CompanyAccountEntity) request.getSession()
+                .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
+        if (transaction != null && companyAccountEntity != null) {
+            String companyNumber = transaction.getCompanyNumber();
+            String smallFullId = smallFullService.generateID(companyNumber);
```
jsimons1989: How will this work once a company has provided multiple accounts if the small-full ID is generated off the company number?

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,62 @@
...
+            SmallFullEntity smallFull = smallFullService.findById(smallFullId);
+
+            if (smallFull != null) {
+                String companyAccountLink = companyAccountEntity.getData().getLinks()
+                        .get("small_full_accounts");
```
jsimons1989: Can "small_full_accounts" be added to LinkType?

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
sroberts2: Agreed.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+        StringBuilder sb = new StringBuilder("CompanyAccountInterceptor failed on preHandle");
+        if (transaction != null) {
+            sb.append(" for transaction ").append(transaction.toString());
+        }
+        sb.append(".");
+        LOGGER.error(sb.toString());
+        return false;
+    }
```
jsimons1989: What does false return to the user?

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,69 @@
...
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) {
+        Map<String, String> pathVariables = (Map) request
+                .getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+        String companyAccountId = pathVariables.get("companyAccountId");
+        CompanyAccountEntity companyAccountEntity = companyAccountService
+                .findById(companyAccountId);
+        Transaction transaction = (Transaction) request.getSession()
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        HttpSession session = request.getSession();
```
jsimons1989: Can you add a line break/break this up for readability please?

sroberts2: Done.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
sroberts2: At the moment nothing it's only there to get the code to compile, once the Global error handling is in place I will look to address this possibly throwing an exception that spawns either a 400 or 404 depending what failed in the interceptor.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+
+            if (accountsLink.equals(accountsSelf)) {
+                session.setAttribute(AttributeName.COMPANY_ACCOUNT.getValue(),
+                        companyAccountEntity);
+                return true;
+            }
+        }
+        StringBuilder sb = new StringBuilder("CompanyAccountInterceptor failed on preHandle");
+        if (transaction != null) {
+            sb.append(" for transaction ").append(transaction.toString());
```
zxchris: This it too complicated, and not very explicit. The result will be if it says `for transaction` the log reader will know the transaction failed, and if it doesn't say `for transaction` the _assumption_ will be that the company-account read failed.

Validate and error at the point of getting a value, not deferred to a later stage. This will avoid you having to wrap a big `if( successful) { ` around the main logic ;-)

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+                if (companyAccountLink.equals(smallFullSelf)) {
+                    session.setAttribute(AttributeName.SMALLFULL.getValue(), smallFull);
+                    return true;
+                }
+            }
+        }
+        StringBuilder sb = new StringBuilder("SmallFullInterceptor failed on preHandle");
+        if (transaction != null) {
+            sb.append(" for transaction ").append(transaction.toString());
+        }
```
zxchris: See `companyAccountsInterceptor` comments re error handling and reducing the number of `if` statements.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+        String companyAccountId = pathVariables.get("companyAccountId");
+
+        CompanyAccountEntity companyAccountEntity = companyAccountService
+                .findById(companyAccountId);
+        Transaction transaction = (Transaction) request.getSession()
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        HttpSession session = request.getSession();
+
+        if (transaction != null && companyAccountEntity != null) {
+            String accountsLink = transaction.getLinks().get(LinkType.COMPANY_ACCOUNT.getLink());
```
zxchris: The link to the `company_accounts` is not in the `links` sub-doc, but is in the `resources: {}` sub-doc.

Note, however that that there could be _multiple_ `company_accounts` resources in a transaction. You will have to iterate through the `resource` sub-doc, filtering on `kind` and checking `resources.<key>.links.resource` is one of the resources.

*Warning* Do not use the `<key>` of the `resources` sub-doc, as that is not a data value and will be imminently hashed.

Hope that makes sense!

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -40,34 +50,52 @@
...
-            String accountsSelf = companyAccountEntity.getData().getLinks()
-                    .get(LinkType.SELF.getLink());
+        String accountsSelf = companyAccountEntity.getData().getLinks()
+                .get(LinkType.SELF.getLink());
+        Map<String, Resources> resourcesList = transaction.getResources();
+        if (!isLinkInResourceMap(resourcesList, accountsSelf)) {
+            LOGGER.error(
+                    "CompanyAccountInterceptor failed on preHandle: Failed to find the CompanyAccount self link in the Transactions links"
+                            + transaction
+                            .toString() + ".");
```
jsimons1989: A transaction will grow very big with all of the filings, resources, etc. My opinion is only the necessary data from the transaction should be logged.

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -63,6 +63,6 @@ public void addInterceptors(final InterceptorRegistry registry) {
         registry.addInterceptor(companyAccountInterceptor)
                 .addPathPatterns("/transactions/{transactionId}/company-accounts/{companyAccountId}/**");
         registry.addInterceptor(smallFullInterceptor)
-                .addPathPatterns("/transactions/{transactionId}/company-accounts/{companyAccountId}/small-full/**");
+                .addPathPatterns("/transactions/{transactionId}/company-accounts/{companyAccountId}/small-full/*");
```
jsimons1989: Will this catch longer URLs?
e.g. /transactions/{transactionId}/company-accounts/{companyAccountId}/small-full/notes/loans-to-directors/loans

jsimons1989:
Unit tests are failing: http://ch-jenkins-2k8.internal.ch:8080/jenkins/job/Build/job/chs/job/company-accounts.api.ch.gov.uk/job/feature/221/

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -1,57 +1,71 @@
...
+        SpringApplication.run(CompanyAccountsApplication.class, args);
+    }
+
+    @Override
+    public void addInterceptors(final InterceptorRegistry registry) {
+        registry.addInterceptor(transactionInterceptor)
+                .addPathPatterns("/transactions/{transactionId}/**");
+        registry.addInterceptor(companyAccountInterceptor)
+                .addPathPatterns(
+                        "/transactions/{transactionId}/company-accounts/{companyAccountId}/**");
```
macreman: This should match `/transactions/{transactionId}/company-accounts/{companyAccountId}**` - notice the last slash has been removed.

This is to ensure that this interceptor is run against the GET, PUT and DELETE methods (not POST which doesn't have the companyAccountId) of `company-accounts` i.e. the check that ensures that the company-accounts resource is on the transaction referenced in the URL

This also gives us the GET implementation for free as if it passes the interceptor we will already have the company-accounts data in the session so the controller can just return that data from the session

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -1,57 +1,71 @@
...
+        registry.addInterceptor(smallFullInterceptor)
+                .addPathPatterns(
+                        "/transactions/{transactionId}/company-accounts/{companyAccountId}/small-full/**");
```
macreman: Very similar comment to the one above for company-accounts, we should remove the last `/` here.

Once slight difference is that this will catch the POST so we'll have to explicitly check for that condition in the interceptor code to allow that through i.e. if HTTP method is POST and path is `/transactions/{transactionId}/company-accounts/{companyAccountId}/small-full` skip the checks

src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
```diff
@@ -26,9 +26,10 @@
     public ResponseEntity create(@Valid @RequestBody CurrentPeriod currentPeriod,
             HttpServletRequest request)
             throws NoSuchAlgorithmException {
-        Transaction transaction = (Transaction) request.getSession().getAttribute(AttributeName.TRANSACTION.getValue());
+        CompanyAccountEntity companyAccountEntity = (CompanyAccountEntity) request.getSession()
+                .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
         CurrentPeriod result = currentPeriodService
-                .save(currentPeriod, transaction.getCompanyNumber());
+                .save(currentPeriod, companyAccountEntity.getId());
```
macreman: similar to https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/18/files we should be checking for error conditions here while writing to the database.

This PR has been merged into develop so the 'ResponseObject' pattern seen there should be used when calling a service from a controller.

src/main/java/uk/gov/companieshouse/api/accounts/controller/SmallFullController.java
```diff
@@ -26,10 +26,9 @@
     public ResponseEntity create(@Valid @RequestBody SmallFull smallFull,
             HttpServletRequest request)
             throws NoSuchAlgorithmException {
-        Transaction transaction = (Transaction) request.getSession()
-                .getAttribute(AttributeName.TRANSACTION.getValue());
-
-        SmallFull result = smallFullService.save(smallFull, transaction.getCompanyNumber());
+        CompanyAccountEntity companyAccountEntity = (CompanyAccountEntity) request.getSession()
+                .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
+        SmallFull result = smallFullService.save(smallFull, companyAccountEntity.getId());
```
macreman: Same as previous comment - we should use the 'ResponseObject' pattern for the service to check for errors and respond accordingly with the correct code in the controller

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,101 @@
...
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) {
+        HttpSession session = request.getSession();
+        Transaction transaction = (Transaction) session
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        if (transaction == null) {
+            LOGGER.error(
```
macreman: use `errorRequest` so we can track request ID

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,101 @@
...
+            return false;
+        }
+
+        Map<String, String> pathVariables = (Map) request
+                .getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+        String companyAccountId = pathVariables.get("companyAccountId");
+        CompanyAccountEntity companyAccountEntity = companyAccountService
+                .findById(companyAccountId);
+        if (companyAccountEntity == null) {
+            LOGGER.error(
```
macreman: use `errorRequest` so we can track request ID

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,101 @@
...
+                            + transaction
+                            .getId() + ".");
+            return false;
+        }
+
+        String accountsSelf = companyAccountEntity.getData().getLinks()
+                .get(LinkType.SELF.getLink());
+        Map<String, Resources> resourcesList = transaction.getResources();
+        if (!isLinkInResourceMap(resourcesList, accountsSelf)) {
+            LOGGER.error(
```
macreman: use `errorRequest` so we can track request ID

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,101 @@
...
+    @SuppressWarnings("unchecked")
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) {
+        HttpSession session = request.getSession();
+        Transaction transaction = (Transaction) session
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        if (transaction == null) {
+            LOGGER.error(
+                    "CompanyAccountInterceptor failed on preHandle: Failed to retrieve a transaction from the session.");
+            return false;
```
macreman: each return false case should set the correct status code to report back to the end user, I'd expect this case to be a 500 as the request should not get to this interceptor without a transaction being stored in the session from the previous interceptor

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,101 @@
...
+                .getAttribute(HandlerMapping.URI_TEMPLATE_VARIABLES_ATTRIBUTE);
+        String companyAccountId = pathVariables.get("companyAccountId");
+        CompanyAccountEntity companyAccountEntity = companyAccountService
+                .findById(companyAccountId);
+        if (companyAccountEntity == null) {
+            LOGGER.error(
+                    "CompanyAccountInterceptor failed on preHandle: Failed to retrieve a CompanyAccount for "
+                            + transaction
+                            .getId() + ".");
+            return false;
```
macreman: each return false case should set the correct status code to report back to the end user, I'd expect this case to be a 404 as the company account with this id is not found. 

We should also handle the database read with a try catch so we can be more explicit with response codes e.g. if we get a null response does that mean the company account doesn't exist so 404 versus catching a MongoException which is a connection issue (or similar) so would be a 500

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,101 @@
...
+
+        String accountsSelf = companyAccountEntity.getData().getLinks()
+                .get(LinkType.SELF.getLink());
+        Map<String, Resources> resourcesList = transaction.getResources();
+        if (!isLinkInResourceMap(resourcesList, accountsSelf)) {
+            LOGGER.error(
+                    "CompanyAccountInterceptor failed on preHandle: Failed to find the CompanyAccount self link in the Transactions links "
+                            + transaction
+                            .getId() + ".");
+            return false;
```
macreman: each return false case should set the correct status code to report back to the end user, I'd expect this case to be a 404 as the company accounts resource is not found for this transaction

Brain dump:
thought process for the 404 vs some other response code is that they get the same response code if the company account resource doesn't exist at all and if it doesn't exist on the correct transaction meaning that if this gets called with a company account id that exists on a different transaction it doesn't give that fact away to the end user, they just know that it doesn't exist rather than them knowing that it does exist but for a different transaction

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) throws NoSuchAlgorithmException {
+
+        HttpSession session = request.getSession();
+        CompanyAccountEntity companyAccountEntity = (CompanyAccountEntity) session
+                .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
+        if (companyAccountEntity == null) {
+            LOGGER.error(
```
macreman: use errorRequest

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+                    "SmallFullInterceptor failed on preHandle: Failed to retrieve a companyAccountEntity from the session.");
+            return false;
```
macreman: set response code before return false - 500 here

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        String companyAccountId = companyAccountEntity.getId();
+        String smallFullId = smallFullService.generateID(companyAccountId);
+        SmallFullEntity smallFull = smallFullService.findById(smallFullId);
+        if (smallFull == null) {
+            LOGGER.error(
```
macreman: use errorRequest

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+                    "SmallFullInterceptor failed on preHandle: Failed to retrieve a SmallFull account from the Database.");
+            return false;
```
macreman: set response code before return false - 500 here if we catch a MongoException or 404 if null is returned?

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        String companyAccountLink = companyAccountEntity.getData().getLinks()
+                .get(LinkType.SMALL_FULL.getLink());
+        String smallFullSelf = smallFull.getData().getLinks().get(LinkType.SELF.getLink());
+        if (!companyAccountLink.equals(smallFullSelf)) {
+            LOGGER.error(
```
macreman: use errorRequest

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,79 @@
...
+                    "SmallFullInterceptor failed on preHandle: The SmallFull self link does not exist in the CompanyAccounts links");
+            return false;
```
macreman: set response code before return false - 500 here? - as the data is in an inconsistent state so not user error (no Id provided for the small-full level so not their fault)

PR #25, opened 2018-08-01
Feature/add model add json properties

ch-mcabrera:
Adding json propery to the ixbrl models to marshal the objects to json with the correct name. 
Adding `Company` model 

Resolves:  SFA-122

_**Relevant Links**_
[Filing Generator Design]( https://companieshouse.atlassian.net/wiki/x/gYAtGg)
[XBRL Model](https://companieshouse.atlassian.net/wiki/x/aQBjGg)

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/ixbrl/Account.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/ixbrl/Account.java
@@ -8,10 +8,13 @@ import uk.gov.companieshouse.api.accounts.model.ixbrl.period.Period;
 
 public class Account {
 
+    @JsonProperty("period")
     private Period period;
     @JsonProperty("balance_sheet")
     private BalanceSheet balanceSheet;
+    @JsonProperty("notes")
     private Notes notes;
+    @JsonProperty("company")
     private Company company;
 
     public Period getPeriod() {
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/model/ixbrl/period/Period.java
```diff
@@ -1,10 +1,16 @@
...
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class Period {
 
+    @JsonProperty("current_period_start_on")
     private String currentPeriodStartOn;
+    @JsonProperty("current_period_ends_on")
     private String currentPeriodEndsOn;
+    @JsonProperty("previous_period_start_n")
```
jsimons1989: typo on "previous_period_start_n"

src/main/java/uk/gov/companieshouse/api/accounts/model/ixbrl/Account.java
```diff
@@ -1,14 +1,18 @@
...
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import uk.gov.companieshouse.api.accounts.model.ixbrl.balancesheet.BalanceSheet;
+import uk.gov.companieshouse.api.accounts.model.ixbrl.company.Company;
 import uk.gov.companieshouse.api.accounts.model.ixbrl.notes.Notes;
 import uk.gov.companieshouse.api.accounts.model.ixbrl.period.Period;
 
 public class Account {
 
     private Period period;
```
cburns93: Could we add the `@JsonProperty` to the other fields as well, this is just to safeguard us against problems in the future if someone was to change the name of the fields as by default they are named after the variable itself. So for example if someone was to change `period` to `periodNew` it would break. We had this problem when renaming the models for the API.

ch-mcabrera: good point. I had thought the same but I didnt changed ;)

PR #34, opened 2018-08-07
Adding company account interceptor plus tests.

sroberts2:
Added the CompanyAccountInterceptor.
Added findById method to CompanyAccount service.

Helps resolve: SFA-509

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
@@ -1,0 +1,130 @@
+package uk.gov.companieshouse.api.accounts.interceptor;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DataAccessException;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.HandlerMapping;
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
+import uk.gov.companieshouse.api.accounts.AttributeName;
+import uk.gov.companieshouse.api.accounts.Kind;
+import uk.gov.companieshouse.api.accounts.LinkType;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
+import uk.gov.companieshouse.api.accounts.service.CompanyAccountService;
+import uk.gov.companieshouse.api.accounts.transaction.Resources;
+import uk.gov.companieshouse.api.accounts.transaction.Transaction;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+
+/**
+ * The CompanyAccountInterceptor interceptor validates that the Company account is in the correct
+ * state for the calling request. It insures a Company account resource exists against the
+ * transaction, checks the account ID is the same as the one that is referenced in the relevant
+ * Transaction. This interceptor will fail if the Transaction interceptor has not run prior to its
+ * own execution.
+ */
+@Component
+public class CompanyAccountInterceptor extends HandlerInterceptorAdapter {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger("company-accounts.api.ch.gov.uk");
+    @Autowired
+    private CompanyAccountService companyAccountService;
+
+    /**
+     * This method extracts the 'company_account' parameter passed via the URI then validates it.
+     * Validation is carried out via a database lookup for a CompanyAccountEntity Object then
+     * matching the retrieved Entities 'self' link to the session stored Transactions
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+            Object handler) {
+
+        final Map<String, Object> debugMap = new HashMap<String, Object>();
+        debugMap.put("request_method", request.getMethod());
+
+        HttpSession session = request.getSession();
+        Transaction transaction = (Transaction) session
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        if (transaction == null) {
+            LOGGER.debugRequest(request,
```
macreman: errorRequest for internal server error case

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,130 @@
...
+ * The CompanyAccountInterceptor interceptor validates that the Company account is in the correct
+ * state for the calling request. It insures a Company account resource exists against the
+ * transaction, checks the account ID is the same as the one that is referenced in the relevant
+ * Transaction. This interceptor will fail if the Transaction interceptor has not run prior to its
+ * own execution.
+ */
+@Component
+public class CompanyAccountInterceptor extends HandlerInterceptorAdapter {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger("company-accounts.api.ch.gov.uk");
```
cburns93: Use `APPLICATION_NAME_SPACE` static variable in CompanyAccountsApplication.java?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -26,28 +27,27 @@
 @Service
 public class CompanyAccountServiceImpl implements CompanyAccountService {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger("company-accounts.api.ch.gov.uk");
```
cburns93: same here

src/test/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptorTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.web.servlet.HandlerMapping;
+import uk.gov.companieshouse.api.accounts.AttributeName;
+import uk.gov.companieshouse.api.accounts.Kind;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountDataEntity;
+import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
+import uk.gov.companieshouse.api.accounts.service.CompanyAccountService;
+import uk.gov.companieshouse.api.accounts.transaction.Resources;
+import uk.gov.companieshouse.api.accounts.transaction.Transaction;
+@ExtendWith(MockitoExtension.class)
```
cburns93: no whitespace between any of these lines? best to add it for separation

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
sroberts2: For a LOGGER.errorRequest the method header is: 
public void errorRequest(HttpServletRequest request, Exception e, Map<String, Object> data).  

I haven't got access to an Exception to pass it what do you suggest? Incidentally I was copying the TransactionAuthenticationInterceptor in its pattern of:

if (authorisedUser == null) { 
LOGGER.debugRequest(request, "TransactionAuthenticationInterceptor error: no authorised identity", null);
response.setStatus(HttpServletResponse.SC_INTERNAL_SERVER_ERROR);
return false;
}
 
We should change them in Tandem given we find a solution.

macreman: Good catch in transaction api case. As discussed that should be using errorRequest too and we can pass null in and the library won't blow up so I'd go for that

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/CompanyAccountInterceptor.java
```diff
@@ -0,0 +1,131 @@
...
+            Object handler) {
+
+        final Map<String, Object> debugMap = new HashMap<String, Object>();
+        debugMap.put("request_method", request.getMethod());
+
+        HttpSession session = request.getSession();
+        Transaction transaction = (Transaction) session
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+        if (transaction == null) {
+            LOGGER.errorRequest(request, null, debugMap);
```
macreman: Looking at this in a little more detail, this is fine to do but gives no "message" field in the logs so I think we need to be explicitly setting that in the line above i.e.
`debugMap.put("message", "CompanyAccountInterceptor error: no transaction in request session");`

PR #35, opened 2018-08-08
Add interceptor for small-full.

sroberts2:
Added the findById method to the AbstractService and its Implementation.
Added the SmallFullInterceptor.

helps resolve: SFA-509

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
@@ -53,6 +53,8 @@ public class SmallFullInterceptor extends HandlerInterceptorAdapter {
     public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
             Object handler) throws NoSuchAlgorithmException {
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/interceptor/SmallFullInterceptor.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param handler - chosen handler to execute, for type and/or instance evaluation
+     * @return true if the execution chain should proceed with the next interceptor or the handler
+     * itself.
+     */
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+            Object handler) throws NoSuchAlgorithmException {
+
+        if (request.getMethod().equalsIgnoreCase("POST") && request.getRequestURI()
```
macreman: probably worth adding a comment on this case to explain this it's to avoid trying to get a small-full resource when the request is trying to create one

sroberts2: Done.

PR #36, opened 2018-08-09
Feature/add filing generator

ch-mcabrera:
Changes to implement the existing **Filing Generator** functionality from _abridged_ to _company-accounts.api_.  Changes: 
- Adding the Filing Generator Controller. Junit tests.
- Adding the Filing Generator Implementation. Junit Tests
Bear in mind that amongst these changes there is an interface and its implementation that contain hardcoded values since I am waiting for the API functionality
 to be there. These files wouldn't need to be reviewed as they HAVE to change. 
  - **AccountsBuilder.java**
  - **SmallFullAccountsBuilderImpl.java**

- Adding the Document Render Service call functionality. 
This Functionality will HAVE to be removed by a SDK call when functionality is there. This has been copied from abridged as it is pointless to reinvent the wheel since it will change. 
You can review them but bear in mind this is existing code and it has to change. Files: 
    - **IxbrlGenerator.java**
    - **IxbrlGeneratorImpl.java**
    - **DocumentGeneratorConnection.java**

Junit tests for the document render service have not been created as it uses static methods and for that reason has been place in a new class so that i could mock it.

Resolves: SFA-122


src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
```diff
@@ -0,0 +1,91 @@
+package uk.gov.companieshouse.api.accounts;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * An enumeration of account types
+ */
+public enum AccountsType {
```
ChrisJBurns: This has been duplicated in the `document-generator-accounts` is this supposed to be duplicated?

src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
```diff
@@ -0,0 +1,43 @@
...
+    @GetMapping("/private/transactions/{transactionId}/company-accounts/{accountId}/filings")
+    public ResponseEntity generateFiling(@PathParam("transactionId") String transactionId,
+        @PathParam("accountId") String accountId, HttpServletRequest request) {
+
+        Transaction transaction = (Transaction)request.getSession().getAttribute(AttributeName.TRANSACTION.getValue());
+
+        try {
+            Filing filing = filingService.generateAccountFiling(transaction, accountId);
+            if (filing != null) {
+                List<Filing> filings = Arrays.asList(filing);
```
ChrisJBurns: I think the initialisation of the `filings` list here is unnecessarily costly. You could just do `return new ResponseEntity<>(Arrays.asList(filing), HttpStatus.OK);`

src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
```diff
@@ -0,0 +1,91 @@
...
+
+        return null;
+    }
+    
+    /**
+     * Method checks if the accountType is present in AccountType Enum
+     * @param accountType
+     */
+    public static boolean isValidAccountType(String accountType) {
+        for (AccountsType type : AccountsType.values()) {
```
ChrisJBurns: This can be simplified using streams I believe :)
```java
    public static boolean isValidAccountType(String accountType) {
        return Arrays.stream(AccountsType.values())
                     .anyMatch(e -> e.equals(accountType));
    }

File: src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+import java.util.Set;
+
+/**
+ * An enumeration of account types
+ */
+public enum AccountsType {
+
+    SMALL_FULL_ACCOUNTS("small-full", "small-full-accounts.html", "accounts#smallfull", "small_full_accounts", "small-full-accounts"),
+    ABRIDGED_ACCOUNTS("abridged", "abridged-accounts.html", "accounts#abridged", "abridged_accounts", "abridged-accounts"),
+    DORMANT_ACCOUNTS("dormant", null, "accounts#dormant", "dormant_accounts", "dormant-accounts"),
```
BSaunders95: Should we treat `null` as a dormant account?

ch-mcabrera: This was copied from abridged and i didnt want to delete the other accounts type to have them as a reference; but i think i will delete them so that it wont cause any confusion.

src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
ch-mcabrera: i copied from abridged and i havent refactored it. But it is good point. i can give it a try

src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
ch-mcabrera: I believe it is supported as this is existing functionality and i copied the file from abridged  and it is working.

src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
```diff
@@ -21,7 +21,7 @@
     @Autowired
     private FilingService filingService;
 
-    @GetMapping("/private/transactions/{transactionId}/company-accounts/{accountId}/filings")
+    @GetMapping("/private/transactions/{transactionId}/company-accounts/{companyAccountId}/filings")
     public ResponseEntity generateFiling(@PathParam("transactionId") String transactionId,
         @PathParam("accountId") String accountId, HttpServletRequest request) {
```
ChrisJBurns: This will also need to change from `accountId` to `companyAccountId`

src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
```diff
@@ -0,0 +1,45 @@
...
+
+@RestController
+public class FilingController {
+
+    @Autowired
+    private FilingService filingService;
+
+    @GetMapping("/private/transactions/{transactionId}/company-accounts/{companyAccountId}/filings")
+    public ResponseEntity generateFiling(@PathParam("transactionId") String transactionId,
+        @PathParam("accountId") String accountId, HttpServletRequest request) {
```
ChrisJBurns: `accountId` should be `companyAccountId` as well as the variable name?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+    private static final String PERIOD_END_ON = "period_end_on";
+
+    @Autowired
+    private ObjectMapper objectMapper;
+    @Autowired
+    private IxbrlGenerator ixbrlGenerator;
+    @Autowired
+    private AccountsBuilder accountsBuilder;
+
+    private EnvironmentReader environmentReader;
```
ChrisJBurns: I would make this variable static and initialise it so it only gets initialised once.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+import uk.gov.companieshouse.api.accounts.util.ixbrl.ixbrlgenerator.DocumentGeneratorConnection;
+import uk.gov.companieshouse.api.accounts.util.ixbrl.ixbrlgenerator.IxbrlGenerator;
+import uk.gov.companieshouse.document.data.DocumentDescriptionHelper;
+import uk.gov.companieshouse.environment.EnvironmentReader;
+import uk.gov.companieshouse.environment.impl.EnvironmentReaderImpl;
+
+@Service
+public class FilingServiceImpl implements FilingService {
+
+    private static final String API_KEY_ENV_VAR = "CHS_API_KEY";
```
ChrisJBurns: I would use the environment reader here like we have used in the Mongo properties, so the variable is loaded once when the class is initialised

ChrisJBurns: Additionally, I would do the same on all other env variables

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Generate the filing for the account type passed in.
+     *
+     * @param transaction - transaction id
+     * @param accountsType - Account type information: account type, ixbrl's template name, account
+     * @return {@link Filing} - null or filing with the filing information (e.g. ixbrl location,
+     * accounts name, etc)
+     * @throws IOException -
+     */
+    private Filing generateAccountFiling(Transaction transaction, AccountsType accountsType)
```
ChrisJBurns: I would make `AccountType` a `Function<AccountType>` so you can use it with a lambda, this should get rid of the need for the `getAccountType` method.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,383 @@
...
+            DOCUMENT_RENDER_SERVICE_END_POINT;
+    }
+
+    /**
+     * Get environment variable value for the passed variable.
+     *
+     * @return environment variable value.
+     */
+    private String getMandatoryEnvVariable(String envVariable) {
+        if (environmentReader == null) {
```
ChrisJBurns: With comment above, you won't need this method anymore

src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
ch-mcabrera: it is not needed, I tested but i will change it to make it consistent.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: I will leave it as it is. but it is good shout

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: Unfortunately, I wont be able to mock it if I change this to be a static variable. I have also tried to autowire it but it didn't work as the Environment Reader hasn't got the component annotation.

ChrisJBurns: I don't think we will be able to mock it at all if it isn't autowired into the context

ch-mcabrera: I've managed to `@Autowired` the EnvironmentReader by creating a bean. And  i now can mock it.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: I wont add load the environment variables into member variables when the class is loaded as I have junit issues if I use this approach.  Therefore, I will get the environmnet variable value only when it is needed.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: this will be removed after Autowiring the Environment reader.

ch-mcabrera:
@BSaunders95 has approved this by mistake. 
Can anyone have a look at it?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Filing generateAccountFiling(Transaction transaction, CompanyAccountEntity accountEntity)
+        throws IOException {
+        Filing filing = null;
+
+        if (isSmallFullAccount(accountEntity)) {
+            filing = generateAccountFiling(transaction,
+                getAccountType(AccountsType.SMALL_FULL_ACCOUNTS.getAccountType()));
```
jterrych: At this point you already know what account type you want. A small full. So using getAccountType is redundant.
```
filing = generateAccountFiling(transaction, AccountsType.SMALL_FULL_ACCOUNTS);
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,387 @@
...
+
+    /**
+     * get the location the document render service needs to stored the ixbrl document. e.g.
+     * "s3://dev-pdf-bucket/chs-dev/accounts/small_full_accounts"
+     *
+     * @param accountsType - Account type information: account type, ixbrl's template name, account
+     * @return
+     */
+    private String getIXBRLLocation(AccountsType accountsType) {
+        if (isSmallFullType(accountsType.getAccountType())) {
```
jterrych: All relevant information about the account type is contained within the enum, and at this point you already have the account type so providing a further check to see if its a small-full doesnt seem necessary.

ch-mcabrera: I added it since as I am thinking to future proof it; depending on the Account Type you passed in the ixbrl could be different. e.g. one for abridged, one for account, etc. 

but this is something tghta this could be added in the future  once **abridged** is migrated.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: Good point ;)

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: Changes for pr's comments: dffd958

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: Changes for pr's comments: dffd958

pom.xml
```diff
@@ -105,12 +106,17 @@
...
 		</dependency>
-
 		<dependency>
 			<groupId>uk.gov.companieshouse</groupId>
 			<artifactId>company-accounts-library</artifactId>
 			<version>${company-accounts-library.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>uk.gov.companieshouse</groupId>
+			<artifactId>document-generator-core</artifactId>
```
jsimons1989: Which version of the document-generator-core is this dependent on? The old library is being replaced by a submodule in [document-generator](https://github.com/companieshouse/document-generator). It won't function as a library anymore as the architecture is changing to a plug-in based approach.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Filing generateAccountFiling(Transaction transaction, CompanyAccountEntity accountEntity)
+        throws IOException {
+        Filing filing = null;
+
+        if (isSmallFullAccount(accountEntity)) {
+            filing = generateAccountFiling(transaction, AccountsType.SMALL_FULL_ACCOUNTS);
+        }
```
jsimons1989: This is going to grow into a big if statement when other account types are added. Can it be refactored into something like this:
```java
AccountsType type = getAccountType(accountEntity);
return generateAccountFiling(transaction, accountType);
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+    private Filing generateAccountFiling(Transaction transaction, AccountsType accountsType)
+        throws IOException {
+        Filing filing = null;
+        Object accountObj = getAccountInformation(accountsType.getAccountType());
+
+        if (accountObj != null) {
+            String ixbrlLocation = callIxbrlGenerator(accountsType,
+                generateJson(accountsType.getAccountType(), accountObj));
+
+            if (ixbrlLocation != null && isValidIXBL()) {
```
jsimons1989: Typo on "isValidIXBL"

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /**
+     * Generates the json for small full account.
+     *
+     * @param account - small full's account information.
+     * @return The account marshaled to JSON and converted to String.
+     * @throws JsonProcessingException
+     */
+    private String generateCompanyAccountJSON(Account account)
```
jsimons1989: This should be called generateSmallFullAccountJSON to match the specific small-full functionality.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+     * @param accountObj - The account's type object that's being processed (e.g. Account)
+     * @return null or ixbrl location
+     * @throws JsonProcessingException
+     */
+    private String generateJson(String accountType, Object accountObj)
+        throws JsonProcessingException {
+        if (isSmallFullType(accountType)) {
+            return generateCompanyAccountJSON((Account) accountObj);
+        }
+        return null;
```
jsimons1989: Logging needs to be added to be able to diagnose issues. This if statement could be missed when future account types are added and logging would help developers quickly find the problem.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+     * the service.
+     *
+     * @param accountType - Account type information (e.g. small-full, abridged, etc)
+     * @param accountObj - The account's type object that's being processed (e.g. Account)
+     * @return null or ixbrl location
+     * @throws JsonProcessingException
+     */
+    private String generateJson(String accountType, Object accountObj)
+        throws JsonProcessingException {
+        if (isSmallFullType(accountType)) {
```
jsimons1989: Instead of:
```java
generateJson(accountsType.getAccountType(), accountObj));

private String generateJson(String accountType, Object accountObj) throws JsonProcessingException {
   if (isSmallFullType(accountType)) {

private boolean isSmallFullType(String filingType) {
        return AccountsType.SMALL_FULL_ACCOUNTS.getAccountType().equals(filingType);
    }
```
it could be simplified to:
```java
generateJson(accountsType, accountObj));

private String generateJson(AccountsType accountType, Object accountObj) throws JsonProcessingException {
   switch(accountType)) {
       case SMALL_FULL:
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /**
+     * Validates the ixbrl against TNEP. This validation is driven by the env. variable and it can
+     * be disable.
+     */
+    private boolean isValidIXBL() {
+
+        boolean isIxbrlValid = true;
+        if ("true".equals(getMandatoryEnvVariable(DISABLE_IXBRL_VALIDATION_ENV_VAR))) {
```
jsimons1989: Can you update the environment reader to get a mandatory boolean instead of doing a String check?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+            //TODO TNEP validation needs to be added. Copy logic from abridged. Next PR
```
jsimons1989: Can you link this to a story instead please? This TODO will be confusing to other developers if something else becomes priority and you don't implement the validation logic next.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+     */
+    private String callIxbrlGenerator(AccountsType accountsType, String requestBody)
+        throws IOException {
+        if (requestBody != null) {
+            DocumentGeneratorConnection connection = getDocumentGeneratorConnection(accountsType,
+                requestBody);
+            return ixbrlGenerator.generateIXBRL(connection);
+        }
+
+        return null;
```
jsimons1989: Can you add logging where you return null please

src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/accountsbuilder/AccountsBuilder.java
```diff
@@ -0,0 +1,13 @@
...
+import java.io.IOException;
+
+/**
+ * Interface to call Accounts API to retrieve the accounts information.
+ *
+ * This functionality has not been implemented yet. Waiting for the API changes to be deployed;
+ * once, that happens the accounts builder implementation needs to be change.
+ **/
+public interface AccountsBuilder {
+    Object buildAccount() throws IOException;
```
jsimons1989: You could use generics here to prevent the need to cast.
```java
public interface AccountsBuilder<T> {
    T buildAccount();
}
```

public class SmallFullAccountsBuilderImpl implements AccountsBuilder<SmallFull> {
    public SmallFull buildAccount() {
    }
}

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+    private final ObjectMapper objectMapper;
+    private final IxbrlGenerator ixbrlGenerator;
+    private final AccountsBuilder accountsBuilder;
+
+    @Autowired
+    public FilingServiceImpl(
+        EnvironmentReader environmentReader,
+        ObjectMapper objectMapper,
+        IxbrlGenerator ixbrlGenerator,
+        AccountsBuilder accountsBuilder) {
```
jsimons1989: Autowiring in the accountsBuilder won't work when you need to support AbridgedAccountsBuilder, SmallFullAccountsBuilder, etc.

src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/accountsbuilder/SmallFullAccountsBuilderImpl.java
```diff
@@ -0,0 +1,80 @@
...
+@Component
+public class SmallFullAccountsBuilderImpl implements AccountsBuilder {
+
+    @Override
+    public Object buildAccount() throws IOException {
+        return getSmallFullAccount();
+    }
+
+    /**
+     * Builds the small full accounts model. Functionality needs to be change.
```
jsimons1989: This comment will be confusing to developers in the future who have not been involved. Please reference Jira tickets explaining the changes.

pom.xml
ch-mcabrera: i will remove the dependency.

src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/accountsbuilder/SmallFullAccountsBuilderImpl.java
```diff
@@ -0,0 +1,80 @@
...
+        account.setNotes(getNotes());
+        account.setCompany(getCompany());
+
+        return account;
+    }
+
+    private Company getCompany() {
+        Company company = new Company();
+        company.setCompanyName("SC344891");
+        company.setCompanyNumber("MYRETON RENEWABLE ENERGY LIMITED");
```
jsimons1989: As part of GDPR we cannot be hardcoding real company data. This should be removed until the company name and number are available.

src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/ixbrlgenerator/DocumentGeneratorConnection.java
```diff
@@ -0,0 +1,94 @@
...
+public class DocumentGeneratorConnection {
+    private String requestMethod;
+    private String serviceURL;
+    private String requestBody;
+    private String authorizationProperty;
+    private String assetId;
+    private String contentType;
+    private String acceptType;
+    private String location;
+    private String templateName;
```
jsimons1989: I wouldn't expect templateName in a DocumentGeneratorConnection object. Can this be split into connect and response objects?

src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/ixbrlgenerator/IxbrlGeneratorImpl.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return The location where the service has stored the generated ixbrl.
+     * @throws IOException
+     */
+    private String callDocumentRenderService(DocumentGeneratorConnection generatorConnection)
+        throws IOException {
+        HttpURLConnection connection = createConnectionForAccount(generatorConnection);
+
+        String requestBody = generatorConnection.getRequestBody();
+        try {
+            if (requestBody != null) {
```
jsimons1989: The request body null check could be before opening the connection to prevent making a connection that won't succeed.

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
```diff
@@ -0,0 +1,461 @@
...
+    /**
+     * Check environmentReaderMock is called the passed-in number of times and the argument's values
+     * of these calls matches the passed-in args.
+     *
+     * @param numOfTimes The number of times environment reader is be called.
+     * @param args the values the environment reader is called with.
+     */
+    private void verifyEnvironmentReaderNumOfCalls(int numOfTimes, String... args) {
+
+        verify(environmentReaderMock, times(numOfTimes))
```
jsimons1989: Instead of needing the int passed in you could simplify to:
```java
private void verifyEnvironmentReaderNumOfCalls(String... args) {
   verify(environmentReaderMock, times(args.length))
```

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    /**
+     * Gets json for small full.
+     *
+     * @return
+     */
+    private String getSmallFullJson() {
+        String json =
+            "{\n"
```
jsimons1989: This will grow massive with all the account types with all the fields. Can the JSON be moved into a file in the test resources folder?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
```diff
@@ -0,0 +1,461 @@
...
+     * @return
+     */
+    private Map<String, String> createAccountEntityLinks() {
+        Map<String, String> dataLinks = new HashMap<>();
+
+        dataLinks.put(LinkType.SELF.getLink(),
+            String.format("/transactions//%s//company-accounts//%s", TRANSACTION_ID, ACCOUNTS_ID));
+
+        dataLinks.put(LinkType.SMALL_FULL.getLink(),
+            String.format("/transactions//%s//company-accounts//%s//small-full//%s",
```
jsimons1989: Why are the double slashes needed? Why won't this work:
```java
String.format("/transactions/%s/company-accounts/%s/small-full/%s", TRANSACTION_ID, ACCOUNTS_ID, SMALL_FULL_ID));
```

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,371 @@
...
+     * @param accountsType - Account type information: account type, ixbrl's template name, account
+     * @param ixbrlLocation - the location where the ixbrl is stored.
+     * @return
+     * @throws IOException
+     */
+    private Filing createAccountFiling(Transaction transaction, AccountsType accountsType,
+        String ixbrlLocation) throws IOException {
+        Filing filing = new Filing();
+
+        //TODO get correct periodEndOn. periodEndOn = Current Period's end date", mongo DB. Waiting for the API changes.
```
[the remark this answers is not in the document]
ch-mcabrera: adf92c4

pom.xml
ch-mcabrera: adf92c4

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: adf92c4

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
ch-mcabrera: adf92c4

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
ch-mcabrera: good shout. 💯

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
ch-mcabrera: 2f26d58

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
ch-mcabrera: 2f26d58

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
ch-mcabrera: 2f26d58

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java
```diff
@@ -0,0 +1,468 @@
...
+
+        Path path = Paths.get(getClass().getClassLoader()
+            .getResource(filePathName).toURI());
+
+        StringBuilder data = new StringBuilder();
+        Stream<String> lines = Files.lines(path);
+        lines.forEach(line -> data.append(line).append("\n"));
+        lines.close();
+
+        return data.toString();
```
jsimons1989: This can be simplified to:
```java
private String getFileContentFromResource(String filePathName) throws URISyntaxException, IOException {
    	Path path = Paths.get(getClass().getClassLoader().getResource(filePathName).toURI());
    	return new String(Files.readAllBytes(path));
}
````

File: src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java (hunk 17 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java]: commit: 4a47945

File: src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java (hunk 14 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java]: 4a47945

File: src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java (hunk 18 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java]: 4a47945

File: src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java (hunk 19 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java]: 4a47945

File: src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java (hunk 21 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java]: 4a47945

File: src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/accountsbuilder/SmallFullAccountsBuilderImpl.java (hunk 24 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/accountsbuilder/SmallFullAccountsBuilderImpl.java]: 4a47945

File: src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java (hunk 32 above)
Comment (ch-mcabrera) [reply src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java]: 💯 Good shout!!!!

Comment (ch-mcabrera) [reply src/test/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImplTest.java]: 9f407eb

File: src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/ixbrlgenerator/IxbrlGeneratorImpl.java (hunk 27 above)
Comment (ch-mcabrera) [reply src/main/java/uk/gov/companieshouse/api/accounts/util/ixbrl/ixbrlgenerator/IxbrlGeneratorImpl.java]: 9f407eb

File: pom.xml [hunk 33]
```diff
@@ -29,10 +29,11 @@
 		<api-helper-java.version>1.0.0-rc1</api-helper-java.version>
 		<structured-logging.version>1.2.0-rc1</structured-logging.version>
 		<aws-java-sdk.version>1.11.52</aws-java-sdk.version>
-		<environment-reader-library.version>1.0.0-rc1</environment-reader-library.version>
-		<company-accounts-library.version>5a6005b</company-accounts-library.version>
+		<environment-reader-library.version>1043eed</environment-reader-library.version>
+		<company-accounts-library.version>9e3b682</company-accounts-library.version>
```
BSaunders95: `company-accounts-library` is now versioned. The latest is `1.0.0-rc2`.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,243 @@
...
+            DOCUMENT_RENDER_SERVICE_END_POINT;
+    }
+
+    /**
+     * Get environment variable value for the passed variable.
+     *
+     * @return environment variable value.
+     */
+    private String getMandatoryEnvVariable(String envVariable) {
+        return environmentReader.getMandatoryString(envVariable);
```
BSaunders95: If a static final EnvironmentReader is declared at the class level, we can declare static strings at the class level to hold environment variables. This means that we don't have to read the same environment variables every time a method is invoked. For example:

```
public class Class {

   private static final EnvironmentReader READER = new EnvironmentReaderImpl();

   private static String DOCUMENT_RENDER_SERVICE_HOST = READER.getMandatoryString("DOCUMENT_RENDER_SERVICE_HOST");

}
```

src/test/java/uk/gov/companieshouse/api/accounts/controller/FilingControllerTest.java
```diff
@@ -0,0 +1,95 @@
...
+    @Mock
+    private HttpSession httpSessionMock;
+    @Mock
+    private HttpServletRequest httpServletRequestMock;
+    @InjectMocks
+    private FilingController filingController;
+
+    @BeforeEach
+    public void setUp() {
+        when(httpServletRequestMock.getAttribute(anyString()))
```
BSaunders95: Can this be split into 2 `when` conditions with specific attribute names? This will make the tests less brittle.

PR #37, opened 2018-08-10
Updating Filing model. Updating jsonproperty name to match others

ch-mcabrera:
Updating Filing model, `links` variable was added by mistake. 
Renaming JsonProperties to match other similar fields so that it passes validation in the `document render service`.

Resolves: SFA-122

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/filing/Filing.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/filing/Filing.java
@@ -20,9 +20,6 @@ public class Filing {
     @JsonProperty("description_values")
     private Map<String, String> descriptionValues;
 
-    @JsonProperty("links")
-    private List<Links> links;
-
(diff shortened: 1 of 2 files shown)


ch-mcabrera:
As far as I know we dont need to add any test for mapping as this json will be validated in other service. 
But I will need to ask @jsimons1989  to confirm this.

ljgeorge:
That's fine. Personally I'd want to know that the message had some chance of being accepted before it left my service, but please discuss with Joanne.

jsimons1989:
I have had a discussion with @ljgeorge. Usually we would test the mapping between the spec model and the model backing the template with the unit tests. As the JSON mappings (JsonProperty annotations) match an iXBRL template we would cover this testing using integration testing (the automated Cucumber tests). We pass our generated documents through a TNEP validator in the filing generator which fails if expected data is missing. The mapping unit tests are missing because the mapping code has not been written yet, only the models exist.

PR #38, opened 2018-08-10
renamed the DB_URL_KEY to TRANSACTIONS_ACCOUNTS_DB_URL

gpatrac:


ChrisJBurns:
Are we sure we want to do these? I specifically changed the name to be accounts because we are trying to keep it generic!

macreman:
@ChrisJBurns this maps directly to the `transactions_accounts` mongo database, if we ever deal with output data for accounts that would be a different database probably called something like `company_accounts` so this should be explicit.

Think of the company profile as an example, we have that already as an output data set under `company_profile`, when we create a company profile api to deal with all company profile data it will have to deal with this output database plus a `transactions_company_profile` database for input data using the transaction model and we'll need to be explicit in mapping those from env vars to properties in code so we should start being explicit now to put this pattern in place

PR #39, opened 2018-08-10
Add handlers for specific error response code

gpatrac:


src/main/java/uk/gov/companieshouse/api/accounts/utility/ApiResponseGenerator.java
```diff
@@ -0,0 +1,27 @@
...
+@Component
+public class ApiResponseGenerator {
+
+    /**
+     * Builds a Response Entity based on the
+     * status, entity and error data encapsulated in the ResponseObject.
+     * @param responseObject - response received from the service methods
+     * @return
+     */
+    public ResponseEntity getApiResponse(ResponseObject responseObject) {
```
macreman: All looks good apart from this method signature, which should take the 3 separate parts of ResponseObject as 3 separate params i.e. getApiResponse(ResponseStatus s, Object o, ErrorData e)

This is so that the controller can control the body returned in a successful case. 
If it does not require a body e.g. for an update/PUT method then it can pass in null for the body to match the spec even if the service layer returns some response object data

PR #45, opened 2018-08-13
Fix to the getById method of the AbstractService class

sroberts2:
Static code analysis raised a concern around type erasure. This branch fixes that, made better use of the Optional and also tightened the Access around the transformer and the repository.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
@@ -40,8 +40,7 @@ public abstract class AbstractServiceImpl<T extends RestObject, U extends BaseEn
 
     @Override
     public U findById(String id) {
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
```diff
@@ -40,11 +40,8 @@ public T save(T rest, String companyAccountId) throws NoSuchAlgorithmException {
 
     @Override
     public U findById(String id) {
-        Optional<U> optional = (Optional<U>) mongoRepository.findById(id);
-        if (optional.isPresent()) {
-            return optional.get();
-        }
-        return null;
+        Optional<U> optional = mongoRepository.findById(id);
```
ChrisJBurns: Pretty sure this can be cut even more?
`return mongoRepository.findById(id).orElse(null);`

PR #46, opened 2018-08-14
Added a null check to the isTransactionIsOpen method

sroberts2:
Introduced a null check into the isTransactionIsOpen method.


src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
```diff
@@ -49,6 +49,7 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
      * Returns whether transaction is open or not.
      */
     private boolean isTransactionIsOpen(ResponseEntity<Transaction> transaction) {
```
ChrisJBurns: Can we instead pass the `Transaction` through rather than the `ResponseEntity<Transaction>`. This makes it more intuitive as the `ResponseEntity` has nothing to do with any of what we're doing in this method. Also worth annotating the method param as `@NotNull` just so it forces people to take care when using the method.

PR #54, opened 2018-08-16
The message digest to be set globally, the application will fail to s…

gpatrac:
the application must fail to start for failures in creating the message digest instance.


src/test/java/uk/gov/companieshouse/api/accounts/service/impl/SmallFullServiceImplTest.java
```diff
@@ -41,7 +46,10 @@
 
         @BeforeEach
         public void setUp() {
+            smallFullService.setMessageDigest(messageDigest);
             when(smallFullTransformer.transform(smallFull)).thenReturn(createdSmallFullEntity);
+            byte[] b = {12, 10, 56, 120, 13, 15};
+            when(messageDigest.digest(any())).thenReturn(b);
         }
 
         @Test
```
sroberts2: Could you remove the throws NoSuchAlgorithmException form the test in this class as the exception is no longer thrown.

PR #57, opened 2018-08-21
Adding private end point to interceptor. Transaction interceptor changes

ch-mcabrera:
Changes to redirect the **Filing Generator**'s requests to a new interceptor which assesses if the transactions is  closed; and to add the private end point as it is needed for the Filing Generator 

Changes 
**1.Code changes has been made to avoid code duplication.**
There are now 3 interceptors handling the transaction
- Transaction Interceptor (existing interceptor) - This interceptor now only gets the transaction from the DB and save it in the request; there will not be any more logic in it.  Transaction status check, which assessed if transaction was open, has been moved to its on interceptor. 
- Open Transaction Interceptor - Get the transaction from request and check if its status is open
- Closed Transaction Interceptor - Get the transaction from request and check if its status is closed; and it is used by the Filing Generator ONly. 
 

**2.Changes to the registry's interceptors to add the new interceptor and updating the companyAccountInterceptor's pattern**
Changes in the CompanyAccountInterceptor to add `{companyAccountId}/**` pattern. To cater for: 
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}`
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}/small-full`
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}/small-full/current-period/`

And it will **NOT re-direct** to the Company Account interceptor the following url: 
- `transactions/111111-111111-111111/company-accounts`

Previous pattern, `{companyAccountId}**, only catered for: 
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}`
And this was causing an error since a small-full request was not going through the interceptor as it needed the CompanyAccountEntity in the request's attribute 

Resolves:  SFA-575

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
@@ -8,10 +8,12 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Primary;
+import org.springframework.core.Ordered;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 import uk.gov.companieshouse.api.accounts.interceptor.CompanyAccountInterceptor;
 import uk.gov.companieshouse.api.accounts.interceptor.SmallFullInterceptor;
+import uk.gov.companieshouse.api.accounts.interceptor.TransactionFilingsInterceptor;
 import uk.gov.companieshouse.api.accounts.interceptor.TransactionInterceptor;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
@@ -24,6 +26,9 @@ public class CompanyAccountsApplication implements WebMvcConfigurer {
 
     @Autowired
     private TransactionInterceptor transactionInterceptor;
+
+    @Autowired
+    private TransactionFilingsInterceptor transactionFilingsInterceptor;
 
     @Autowired
     private CompanyAccountInterceptor companyAccountInterceptor;
@@ -59,14 +64,18 @@ public class CompanyAccountsApplication implements WebMvcConfigurer {
     @Override
     public void addInterceptors(final InterceptorRegistry registry) {
         registry.addInterceptor(transactionInterceptor)
+            .order(Ordered.HIGHEST_PRECEDENCE)
+            .addPathPatterns("/transactions/{transactionId}/**");
+
+        registry.addInterceptor(transactionFilingsInterceptor)
+            .order(Ordered.HIGHEST_PRECEDENCE)
             .addPathPatterns(
-                "/transactions/{transactionId}/**",
-                "/private/transactions/{transactionId}/**");
+                "/private/transactions/{transactionId}/company-accounts/{companyAccountId}/filings");
 
         registry.addInterceptor(companyAccountInterceptor)
             .addPathPatterns(
                 "/transactions/{transactionId}/company-accounts/{companyAccountId}**",
-                "/private/transactions/{transactionId}/company-accounts/{companyAccountId}/filings");
+                "/private/transactions/{transactionId}/company-accounts/{companyAccountId}/**");
 
         registry.addInterceptor(smallFullInterceptor)
             .addPathPatterns(
--- a/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
@@ -14,7 +14,7 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 import uk.gov.companieshouse.api.accounts.interceptor.CompanyAccountInterceptor;
 import uk.gov.companieshouse.api.accounts.interceptor.SmallFullInterceptor;
 import uk.gov.companieshouse.api.accounts.interceptor.ClosedTransactionInterceptor;
-import uk.gov.companieshouse.api.accounts.interceptor.TransactionInterceptor;
+import uk.gov.companieshouse.api.accounts.interceptor.OpenedTransactionInterceptor;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 
@@ -25,7 +25,7 @@ public class CompanyAccountsApplication implements WebMvcConfigurer {
     private static final Logger LOGGER = LoggerFactory.getLogger(APPLICATION_NAME_SPACE);
 
     @Autowired
-    private TransactionInterceptor transactionInterceptor;
+    private OpenedTransactionInterceptor openedTransactionInterceptor;
 
     @Autowired
     private ClosedTransactionInterceptor closedTransactionInterceptor;
@@ -63,7 +63,7 @@ public class CompanyAccountsApplication implements WebMvcConfigurer {
 
     @Override
     public void addInterceptors(final InterceptorRegistry registry) {
-        registry.addInterceptor(transactionInterceptor)
+        registry.addInterceptor(openedTransactionInterceptor)
             .order(Ordered.HIGHEST_PRECEDENCE)
             .addPathPatterns("/transactions/{transactionId}/**");
 
--- a/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
@@ -13,7 +13,7 @@ import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 import uk.gov.companieshouse.api.accounts.interceptor.CompanyAccountInterceptor;
 import uk.gov.companieshouse.api.accounts.interceptor.SmallFullInterceptor;
-import uk.gov.companieshouse.api.accounts.interceptor.TransactionFilingsInterceptor;
+import uk.gov.companieshouse.api.accounts.interceptor.ClosedTransactionInterceptor;
 import uk.gov.companieshouse.api.accounts.interceptor.TransactionInterceptor;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
@@ -28,7 +28,7 @@ public class CompanyAccountsApplication implements WebMvcConfigurer {
     private TransactionInterceptor transactionInterceptor;
 
     @Autowired
-    private TransactionFilingsInterceptor transactionFilingsInterceptor;
+    private ClosedTransactionInterceptor closedTransactionInterceptor;
 
     @Autowired
     private CompanyAccountInterceptor companyAccountInterceptor;
@@ -67,7 +67,7 @@ public class CompanyAccountsApplication implements WebMvcConfigurer {
             .order(Ordered.HIGHEST_PRECEDENCE)
             .addPathPatterns("/transactions/{transactionId}/**");
 
-        registry.addInterceptor(transactionFilingsInterceptor)
+        registry.addInterceptor(closedTransactionInterceptor)
             .order(Ordered.HIGHEST_PRECEDENCE)
             .addPathPatterns(
                 "/private/transactions/{transactionId}/company-accounts/{companyAccountId}/filings");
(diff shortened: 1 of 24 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
```diff
@@ -28,26 +28,63 @@
...
+     * Check transaction is valid. e.g. Accounts api call needs the Transaction to be open. Filing
+     * Generator call needs the transaction to be closed.
+     *
+     * @param request
+     * @param transaction
+     * @return
+     */
+    private boolean isValidTransaction(HttpServletRequest request,
+        ResponseEntity<Transaction> transaction) {
+        if (isFilingGeneratorRequest(request)) {
```
drosser: This method could quickly get out of hand as we add the need to check different transaction statuses for different reasons. Is it worth refactoring this filing logic out into a different interceptor?

Initially, it sounds like the drawback would be that we'd arguably need to get the transaction status again, meaning another potential call to the Transactions API - however, we could add logic in both interceptors (this and the new one) that checks the request to see if it has the transaction added to it already (see line #38 that sets the transaction on the request). If it doesn't have the transaction info, it goes and gets it from the Transactions API, otherwise it just gets the details from the request. 

Alternatively, we could add order precedence when adding the interceptors in `CompanyAccountsApplication` and guarantee that the interceptor that retrieves the transaction is called first (see https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/InterceptorRegistration.html#order-int-), however I think this approach is flawed as it makes assumed knowledge about a different interceptor, tightly coupling them and making them difficult to refactor.

We could also factor out the `isTransactionIsOpen`/`isTransactionClosed` methods into some kind of utility, to help promote re-usability of code.

It feels like this interceptor is too general purpose, and runs the risk of becoming unwieldy. Any kind of separation of responsibility we can promote would benefit us strongly in the long run.

src/test/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptorTest.java
```diff
@@ -31,6 +31,11 @@
 @TestInstance(Lifecycle.PER_CLASS)
 public class TransactionInterceptorTest {
 
+    private static final String ACCOUNTS_ID = "1234561";
+    private static final String ACCOUNTS_API_END_POINT = "http://public/transactions/%s/company-accounts/%s";
+    private static final String FILING_GENERATOR_END_POINT = "http://private/transactions/%s/company-accounts/%s/filings";
+    private static final String TRANSACTION_ID = "1234561-1234561-1234561";
```
drosser: Could these be changed at all?

If you concatenated `ACCOUNTS_ID` and `TRANSACTION_ID` to `ACCOUNTS_API_END_POINT` when declaring the `String` (as opposed to concatenating  on each `get` call), the `FILING_GENERATOR_END_POINT` could become 

`private static final String `FILING_GENERATOR_END_POINT = ACCOUNTS_API_END_POINT + "/filings"'

By doing this, we could also remove the need for the getters.

ch-mcabrera: It is different approach but either way it works. I could change it but i dont see the benefit on it.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
ch-mcabrera: I could agree more with this comment. 
> This method could quickly get out of hand as we add the need to check different transaction statuses for different reasons.

I can see that interceptor was designed to cater for the accounts api calls, but it seems Filing Generator was never taken  into consideration when designing it. It might be worth to check the interceptor's design again to future proof it. 

@drosser, this may be conversation we should have to discuss with @sroberts2 @gpatrac.

ch-mcabrera: Regarding this, 
> Initially, it sounds like the drawback would be that we'd arguably need to get the transaction status again, meaning another potential call to the Transactions API 

I don't see your pont in here: `meaning another potential call to the Transactions API`.
The transaction's information is already retrieved before calling the method, and there will not be any extra calls to the transaction api: 
```JAVA
ResponseEntity<Transaction> transaction = transactionManager
    .getTransaction(transactionId, request.getHeader("X-Request-Id"));
    .getTransaction(transactionId, request.getHeader("X-Request-Id"));
request.setAttribute(AttributeName.TRANSACTION.getValue(), transaction.getBody());
```

ch-mcabrera: Regarding this: 
> however, we could add logic in both interceptors (this and the new one) that checks the request to see if it has the transaction added to it already (see line #38 that sets the transaction on the request). If it doesn't have the transaction info, it goes and gets it from the Transactions API, otherwise it just gets the details from the request.

This is more design issue, and it has been decided to get the transaction at the fist entry point, `TransactionInterceptor`.  We wouldn't need get the `transaction` information in the other interceptors

ch-mcabrera: Regarding this: 
>Alternatively, we could add order precedence when adding the interceptors in CompanyAccountsApplication and guarantee that the interceptor that retrieves the transaction is called first (see https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/InterceptorRegistration.html#order-int-), however I think this approach is flawed as it makes assumed knowledge about a different interceptor, tightly coupling them and making them difficult to refactor.

> We could also factor out the isTransactionIsOpen/isTransactionClosed methods into some kind of utility, to help promote re-usability of code.

> It feels like this interceptor is too general purpose, and runs the risk of becoming unwieldy. Any kind of separation of responsibility we can promote would benefit us strongly in the long run.

This is more design issue, @drosser, this may be conversation we should have to discuss with @sroberts2 @gpatrac.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/TransactionInterceptor.java
```diff
@@ -28,26 +28,64 @@
...
+    /**
+     * Check transaction is valid. e.g. Accounts api call needs the Transaction to be open. Filing
+     * Generator call needs the transaction to be closed.
+     *
+     * @param request
+     * @param transaction
+     * @return
+     */
+    private boolean isValidTransaction(HttpServletRequest request,
+        ResponseEntity<Transaction> transaction) {
```
ChrisJBurns: I would accept a `Transaction` and not the `ResponseEntity`. The method does nothing with the `ReponseEntity` object so you can just pass in the `Transaction` instead.

ch-mcabrera:
Bear in mind that  the private end points have not been discussed yet, see **SFA Low Level API Design**.  However, we need the private end point for the filing generator. 
Any other end points will need to be revisited.

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -58,13 +63,20 @@ public ObjectMapper objectMapper() {
...
+
+        registry.addInterceptor(closedTransactionInterceptor).order(Ordered.HIGHEST_PRECEDENCE)
+            .addPathPatterns(
+                "/private/transactions/{transactionId}/company-accounts/{companyAccountId}/filings");
+
         registry.addInterceptor(companyAccountInterceptor)
-                .addPathPatterns(
-                        "/transactions/{transactionId}/company-accounts/{companyAccountId}**");
+            .addPathPatterns(
+                "/transactions/{transactionId}/company-accounts/{companyAccountId}/**",
```
macreman: the last `/` should not be added before `**` i.e. leave this path as 
`"/transactions/{transactionId}/company-accounts/{companyAccountId}**"`

This change will break the GET/PUT/DELETE methods on the `/transactions/{transactionId}/company-accounts/{companyAccountId}` resource which requires the interceptor to run first to get the data from the DB before proceeding

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -58,13 +63,20 @@ public ObjectMapper objectMapper() {
...
+                "/private/transactions/{transactionId}/company-accounts/{companyAccountId}/**");
```
macreman: same as above, there shouldn't be a `/` before the `**` i.e.
`"/private/transactions/{transactionId}/company-accounts/{companyAccountId}/**"`

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
macreman: it may be worth adding a comment inline here to explain that the trailing `/` before `**` is left out on purpose to avoid this in future

ch-mcabrera: By adding the `/**` 
i.e. ``` /transactions/{transactionId}/company-accounts/{companyAccountId}/**```

it will cater for the following urls' request to be re-directed to the CompanyAccountInterceptor: 
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}`
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}/small-full`
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}/small-full/current-period/`

And it will **NOT re-direct** to the Company Account interceptor the following url: 
- `transactions/111111-111111-111111/company-accounts`


On the other hand, by having only the  `**`
    i.e. `/transactions/{transactionId}/company-accounts/{companyAccountId}**`will cater only for: 
- `transactions/111111-111111-111111/company-accounts/{companyAccountId}`

This means, a request to `transactions/111111-111111-111111/company-accounts/{companyAccountId}/small-full` will not go through the company account interceptor. 
Therefore, it will fail when hitting the SmallFullInterceptor as it needs the CompanyAccountEntity to be in the session.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/ClosedTransactionInterceptor.java
```diff
@@ -1,56 +1,53 @@
...
+     * @param request
      */
-    private boolean isTransactionClosed(ResponseEntity<Transaction> transaction) {
-        return (transaction.getBody() != null && transaction.getBody().getStatus()
-            .equals(TransactionStatus.CLOSED.getStatus()));
+    private void logRequestError(HttpServletRequest request) {
+        final Map<String, Object> debugMap = new HashMap<>();
+        debugMap.put("request_method", request.getMethod());
+        debugMap.put("message",
+            "ClosedTransactionInterceptor error: no transaction in request session");
```
macreman: This message is specific to a certain case so shouldn't be in a generic method, given that this method is only used once and the calling method is now very short it should probably be done inline removing this method entirely

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/OpenTransactionInterceptor.java
```diff
@@ -1,53 +1,54 @@
...
+     * @param request
      */
-    private boolean isTransactionIsOpen(ResponseEntity<Transaction> transaction) {
-        return (transaction.getBody() != null && transaction.getBody().getStatus()
-                .equals(TransactionStatus.OPEN.getStatus()));
+    private void logRequestError(HttpServletRequest request) {
+        final Map<String, Object> debugMap = new HashMap<>();
+        debugMap.put("request_method", request.getMethod());
+        debugMap
+            .put("message", "OpenTransactionInterceptor error: no transaction in request session");
```
macreman: same thing here about a specific error message in a generic method - can again be done inline rather than extracting to a method

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
```diff
@@ -62,18 +71,38 @@ public ObjectMapper objectMapper() {
...
-        registry.addInterceptor(loggingInterceptor)
-                .excludePathPatterns("/healthcheck");
-
-        registry.addInterceptor(transactionInterceptor)
-                .addPathPatterns("/transactions/{transactionId}/**");
 
+        registry.addInterceptor(loggingInterceptor)
+            .excludePathPatterns("/healthcheck");
+
+        registry.addInterceptor(transactionInterceptor).order(Ordered.HIGHEST_PRECEDENCE)
```
macreman: Ordering should be managed by the order that interceptors are registered and the logging interceptor should be left as the highest priority so it catches the entire time for the request.

src/main/java/uk/gov/companieshouse/api/accounts/interceptor/ClosedTransactionInterceptor.java
ch-mcabrera: I thought so. I'll change it

src/main/java/uk/gov/companieshouse/api/accounts/CompanyAccountsApplication.java
ch-mcabrera: my mistake. I fixed the merged issue and I didnt fix the logic.

PR #58, opened 2018-08-22
corrected the URI for the company-account getter

gpatrac:
The     @GetMapping for company-account needs to include companyAccountId.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
@@ -56,7 +56,7 @@ public class CompanyAccountController {
                 .map(result.getStatus(), result.getData(), result.getErrorData());
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -56,7 +56,7 @@ public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount co
                 .map(result.getStatus(), result.getData(), result.getErrorData());
     }
 
-    @GetMapping
+    @GetMapping("/{companyAccountId}")
```
BSaunders95: Can you add this to the `RequestMapping` at the class level please?

drosser: This can't be done as it'll be propagated onto the create, which is invalid

PR #59, opened 2018-08-22
refactored - renamed the service method names to follow companies house naming convention

gpatrac:
SFA-564


BSaunders95:
As part of this work, should we be renaming `get` methods as `retrieve` to follow `CRUD`? Not necessarily a concern, just food for thought

PR #61, opened 2018-08-29
Changed PatchException to extend RuntimeException

sroberts2:
Changed this exception to a runtime exception this will allow the @Transactional spring annotation to work as expected should we need to implement it at a later date. This will also bring it inline with the other errors that can be thrown via the database driver libraries.


macreman:
My understanding after yesterdays meeting was that we would not be doing this and that we should be catching these kinds of exceptions to deal with ourselves

PR #63, opened 2018-08-30
Adding Filing Generator implementation.

ch-mcabrera:
Changes to implement the existing **Filing Generator** functionality from _abridged_ to _company-accounts.api_.  
Changes: 
- Adding empty Filing Generator implementation.
- Updating controller to call the implementation.
- Update and add junit tests

Resolves: SFA-559

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
@@ -1,12 +1,18 @@
 package uk.gov.companieshouse.api.accounts.controller;
 
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.websocket.server.PathParam;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RestController;
 import uk.gov.companieshouse.api.accounts.CompanyAccountsApplication;
 import uk.gov.companieshouse.api.accounts.model.filing.Filing;
+import uk.gov.companieshouse.api.accounts.service.FilingService;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final Logger LOGGER = LoggerFactory
+        .getLogger(CompanyAccountsApplication.APPLICATION_NAME_SPACE);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Filing generateAccountFiling(String transaction, String accountEntity) {
+
+        return null;
```
drosser: By returning null here, we can only ever error. I would return a `new Filing()`, so at least we can always get a happy-path-200 response until we hook in the call to the new document generator API.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+public class FilingServiceImpl implements FilingService {
+
+    private static final Logger LOGGER = LoggerFactory
+        .getLogger(CompanyAccountsApplication.APPLICATION_NAME_SPACE);
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Filing generateAccountFiling(String transaction, String accountEntity) {
```
drosser: We should remove these method variables as they aren't being used yet. Same applies for the interface

src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
```diff
@@ -16,9 +22,23 @@
...
-    public ResponseEntity generateFiling() {
+    public ResponseEntity generateFiling(@PathParam("transactionId") String transactionId,
+        @PathParam("companyAccountId") String accountId, HttpServletRequest request) {
+
+        Filing filing = filingService.generateAccountFiling();
+        if (filing != null) {
+            return new ResponseEntity<>(Arrays.asList(filing), HttpStatus.OK);
+        }
+
+        final Map<String, Object> debugMap = new HashMap<>();
```
ChrisJBurns: We could put this into its own method so other methods could use it as it may get duplicated in the long run? However, I'm not fussy on this, more of a observation for future reference.

ch-mcabrera: it was in a method before in the previous pr. I will move it to separate method

PR #64, opened 2018-08-30
Refactored company account error handling.

sroberts2:
Refactored error handling in company account to match the design decisions made in the recent meeting.

- Created the DataException exception class.
- Changed the Controller to make the decision on what to return from the API calls made.
- Changed the service layer to return an exception where an one has occurred (Wrapped this in the new DataException where applicable).
- Enforced the Error handling for the Controller classes by making the exceptions thrown checked exceptions.
- Updated the Unit test accordingly.

Helps Resolve: SFA-668

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
@@ -2,6 +2,8 @@ package uk.gov.companieshouse.api.accounts.controller;
 
 import static uk.gov.companieshouse.api.accounts.CompanyAccountsApplication.APPLICATION_NAME_SPACE;
 
+import java.util.HashMap;
+import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.validation.Valid;
@@ -15,7 +17,6 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import uk.gov.companieshouse.api.accounts.AttributeName;
-import uk.gov.companieshouse.api.accounts.CompanyAccountsApplication;
 import uk.gov.companieshouse.api.accounts.exception.DataException;
 import uk.gov.companieshouse.api.accounts.exception.PatchException;
 import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
@@ -1,6 +1,5 @@
 package uk.gov.companieshouse.api.accounts.controller;
 
-import com.mongodb.DuplicateKeyException;
 import java.util.HashMap;
 import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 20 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/CompanyAccountService.java
```diff
@@ -1,18 +1,19 @@
...
 import uk.gov.companieshouse.api.accounts.transaction.Transaction;
 
 /**
  * A service for {@link CompanyAccountEntity} and its data {@link CompanyAccountDataEntity}
  */
 public interface CompanyAccountService {
 
-    ResponseObject createCompanyAccount(CompanyAccount companyAccount, Transaction transaction,
-            String requestId);
+    CompanyAccount createCompanyAccount(CompanyAccount companyAccount, Transaction transaction,
```
macreman: this should still be returning a ResponseObject

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -50,10 +53,26 @@ public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount co
...
-                .createCompanyAccount(companyAccount, transaction, requestId);
-        return apiResponseMapper
-                .map(result.getStatus(), result.getData(), result.getErrorData());
+
+        ResponseEntity responseEntity;
+        try {
+            CompanyAccount companyAccountInserted = companyAccountService
+                    .createCompanyAccount(companyAccount, transaction, requestId);
+            responseEntity = apiResponseMapper
+                    .map(ResponseStatus.SUCCESS_CREATED, companyAccountInserted);
```
macreman: we should be using response.getStatus() here rather than having controllers deciding which ResponseStatus to use. 

setting the ResponseStatus is the responsibility of the service, the controller uses this to map to a HTTP response code (via the apiResponseMapper utility)

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -50,10 +53,26 @@ public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount co
...
+        } catch (ApiErrorResponseException aere) {
+            LOGGER.error(aere);
+            responseEntity = apiResponseMapper.map(ResponseStatus.TRANSACTION_PATCH_ERROR);
```
macreman: My understanding from the meeting the other day was that we would no longer use ResponseStatus mappings for exceptional responses (i.e. 5XX errors) so this should explicitly be set as a 500 error in the controller without using the ResponseStatus.

The apiResponseMapper could be changed to map Exception types to http response codes instead so this would become:
`responseEntity = apiResponseMapper.map(aere);`
and the updated response mapper method would map an `ApiErrorResponseException` to a 500

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -50,10 +53,26 @@ public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount co
...
+        } catch (DataException se) {
+            LOGGER.error(se);
+            if (se.getCause() instanceof DuplicateKeyException) {
+                responseEntity = apiResponseMapper.map(ResponseStatus.DUPLICATE_KEY_ERROR);
```
macreman: same comment here - response status should not be used for mapping error responses

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -50,10 +53,26 @@ public ResponseEntity createCompanyAccount(@Valid @RequestBody CompanyAccount co
...
+            } else {
+                responseEntity = apiResponseMapper.map(ResponseStatus.MONGO_ERROR);
```
macreman: same comment here - response status should not be used for mapping error responses

src/main/java/uk/gov/companieshouse/api/accounts/controller/CompanyAccountController.java
```diff
@@ -71,7 +90,8 @@ public ResponseEntity getCompanyAccount(HttpServletRequest request) {
             return ResponseEntity.status(HttpServletResponse.SC_NOT_FOUND).body(null);
         }
 
-        return ResponseEntity.status(HttpStatus.OK).body(companyAccountTransformer.transform(companyAccountEntity));
+        return ResponseEntity.status(HttpStatus.OK)
```
macreman: I know this is only changed for splitting the line but have to point out that this code needs to be updated to handle errors in the transform method i.e. if the transform fails we should detect that and deal with it appropriately with a useful log message and response code

PR #69, opened 2018-09-04
Refactor the abstract service to run the same as company account

sroberts2:
Re-factored the abstract service to have the same error handling process as the CompanyAccounts service, changed the unit tests as a result. This effects both the CurrentPeriod and the SmallFull. Noticed and fixed some formatting issues on the CompanyAccountController.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
@@ -6,7 +6,6 @@ import java.security.NoSuchAlgorithmException;
 import java.util.HashMap;
 import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
 import javax.validation.Valid;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.dao.DataAccessException;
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
```diff
@@ -37,23 +43,32 @@ public AbstractServiceImpl(MongoRepository<U, String> mongoRepository,
...
         U baseEntity = genericTransformer.transform(rest);
 
-        try{
-        baseEntity.setId(generateID(companyAccountId));
+        final Map<String, Object> debugMap = new HashMap<>();
+        debugMap.put("transaction_id", transaction.getId());
+        debugMap.put("company_accounts_id", companyNumber);
+
+        try {
+            baseEntity.setId(generateID(companyNumber));
```
macreman: should be using company accounts id not company number

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
```diff
@@ -37,23 +43,32 @@ public AbstractServiceImpl(MongoRepository<U, String> mongoRepository,
...
+
         addEtag(rest);
         addKind(rest);
         U baseEntity = genericTransformer.transform(rest);
 
-        try{
-        baseEntity.setId(generateID(companyAccountId));
+        final Map<String, Object> debugMap = new HashMap<>();
+        debugMap.put("transaction_id", transaction.getId());
+        debugMap.put("company_accounts_id", companyNumber);
```
macreman: this isn't the company_accounts_id, we shouldn't be using companyNumber here at all

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/AbstractServiceImpl.java
macreman: generateID needs to be different per resource, it is built up from `companyaccountsId` concatentated with the resource type so we'd need the resource type either passed as a parameter or different implementations of generateID per resource

e.g. for a small-full resource we'd use a hash of `{companyAccountId}-small-full` and for a current period we'd use `{companyAccountId}-current-period`

src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
```diff
@@ -44,36 +52,61 @@ public ResponseEntity create(@Valid @RequestBody CurrentPeriod currentPeriod,
...
+        try {
+            responseObject = currentPeriodService.findById(currentPeriodId);
+        } catch (DataAccessException dae) {
+            LOGGER.errorRequest(request, dae, debugMap);
+            return apiResponseMapper.map(dae);
         }
 
-        return ResponseEntity.status(HttpStatus.OK).body(currentPeriodEntity);
+        CurrentPeriod currentPeriod = responseObject.getData();
+        return apiResponseMapper.mapGetResponse(currentPeriod, request);
```
macreman: Since we have the ResponseObject with a ResponseStatus here we should use the regular map method and pass responseObject.getStatus(), getData(), getValidationErrorData()
The map method will need to handle the new ResponseStatus.FOUND/NOT_FOUND values anyway so we should use it when we have a ResponseObject

PR #70, opened 2018-09-04
Feature/filing gen add basic implementation

ch-mcabrera:
Code changes to add a basic implementation of the filing generator. 

The filing generator returns a filing object but empty at the moment if it is a small-full account what we are dealing with. It is an small full account type if the `small-full` link exists in links section within Account's data as suggested by one of the architects. 

The **DocumentGeneratorCaller** class has been added to replace the call to the Doc. Generator's new end point, which is needed to generate the ixbrl and obtain the location that will be set in the returned filing object. Since this functionality is not available yet, this class returns an empty string allowing the filing generator request to be completed. 

Resolves: SFA-559


src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
```diff
@@ -0,0 +1,88 @@
...
+        String filingDescriptionKey) {
+        this.accountType = accountType;
+        this.templateName = templateName;
+        this.kind = kind;
+        this.resourceKey = resourceKey;
+        this.filingDescriptionKey = filingDescriptionKey;
+    }
+
+    public static AccountsType getAccountsType(String accountType) {
+        for (AccountsType accountsType : AccountsType.values()) {
```
BSaunders95: Could this be turned into a stream to make it more performant?

src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
```diff
@@ -0,0 +1,88 @@
...
+            if (accountsType.getAccountType().equals(accountType)) {
+                return accountsType;
+            }
+        }
+
+        return null;
```
BSaunders95: Is this exceptional behaviour? If so, a checked exception should be thrown here (which could become part of the stream by using an `orElseThrow` clause)

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -1,21 +1,99 @@
...
+     */
+    private AccountsType getAccountType(CompanyAccountEntity accountEntity) {
+        Map<String, String> links = accountEntity.getData().getLinks();
+
+        if (links.containsKey(AccountsType.SMALL_FULL_ACCOUNTS.getResourceKey())) {
+            return AccountsType.SMALL_FULL_ACCOUNTS;
+        }
+
+        Map<String, Object> logMap = new HashMap<>();
+        logMap.put(LOG_ERROR_KEY, "Account Type not found");
```
BSaunders95: You don't need this entry in the map; it's captured in the log line below:
```
LOGGER.error("Account Type not found", logMap);
```

src/main/java/uk/gov/companieshouse/api/accounts/utility/ixbrl/DocumentGeneratorCaller.java
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.stereotype.Component;
+
+/**
+ * Class to call the document generator to obtain the ixbrl location.
+ * Since the functionality has not been implemented yet, (STORY SFA-574), it returns an empty string.
+ *
+ * This class will change to call the new end point and it will return the s3 ixbrl location.
+ */
+@Component
+public class DocumentGeneratorCaller {
```
BSaunders95: Could this class implement an interface to make unit testing simpler?

ch-mcabrera: In my opinion, we should be adding an interface for one method only which returns an empty string.

src/main/java/uk/gov/companieshouse/api/accounts/AccountsType.java
ch-mcabrera: By changing this to Functional Programming will not make it more performant. However, I will remove this method as it is not used anyway.

PR #73, opened 2018-09-06
Feature/filing gen create filing model

ch-mcabrera:
Code changes to create the returned filing model. 

The TNEP validation method has also been added so that functionality can be plugged-in when it is completed. This validation always occur as long as it has not been disable by using a environment variable.

The controllor returned error has changed from internal error to bad request as per model. 

Resolves: SFA-559

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
 		<api-helper-java.version>1.0.0-rc1</api-helper-java.version>
 		<structured-logging.version>1.4.0-rc2</structured-logging.version>
 		<aws-java-sdk.version>1.11.52</aws-java-sdk.version>
-		<environment-reader-library.version>1.0.0-rc1</environment-reader-library.version>
+		<environment-reader-library.version>1043eed</environment-reader-library.version>
 		<company-accounts-library.version>5a6005b</company-accounts-library.version>
 		<jackson-datatype-jsr310.version>2.9.6</jackson-datatype-jsr310.version>
 		<httpclient.version>4.5.6</httpclient.version>
--- a/src/main/java/uk/gov/companieshouse/api/accounts/ApplicationConfiguration.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/ApplicationConfiguration.java
@@ -5,6 +5,8 @@ import java.security.NoSuchAlgorithmException;
 import org.apache.commons.codec.digest.MessageDigestAlgorithms;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import uk.gov.companieshouse.environment.EnvironmentReader;
+import uk.gov.companieshouse.environment.impl.EnvironmentReaderImpl;
 
 /**
  * General application configuration .
@@ -15,5 +17,10 @@ public class ApplicationConfiguration {
     @Bean
     public MessageDigest getMessageDigest() throws NoSuchAlgorithmException {
         return MessageDigest.getInstance(MessageDigestAlgorithms.SHA_256);
+    }
+
+    @Bean
+    public EnvironmentReader environmentReader() {
+        return  new EnvironmentReaderImpl();
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
@@ -39,7 +39,7 @@ public class FilingController {
 
         if (transaction == null) {
             logRequestError(request, "no transaction in request session");
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
         }
 
         CompanyAccountEntity companyAccountEntity =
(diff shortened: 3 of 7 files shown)


pom.xml
```diff
@@ -29,10 +29,9 @@
     <api-helper-java.version>1.0.0-rc1</api-helper-java.version>
     <structured-logging.version>1.4.0-rc2</structured-logging.version>
     <aws-java-sdk.version>1.11.52</aws-java-sdk.version>
-    <environment-reader-library.version>1.0.0-rc1</environment-reader-library.version>
+    <environment-reader-library.version>1043eed</environment-reader-library.version>
```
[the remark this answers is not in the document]
ch-mcabrera: We are using the snapshot at the moment as I added a new method in the library

src/main/java/uk/gov/companieshouse/api/accounts/ApplicationConfiguration.java
```diff
@@ -20,14 +19,14 @@
...
-    }
-    
+
     @Bean
     public EnvironmentReader getEnvironmentReader() {
         return new EnvironmentReaderImpl();
     }
+
+    @Bean
+    public RestTemplate getRestTemplate() {
```
ChrisJBurns: So just to be clear we are injecting this the application context to use in constructor injection?

ch-mcabrera: This has been added for the same reason as we added the **EnvironmentReader bean** so that we can autowire it whenever it is needed. 
FYI. RestTemplate has been added for the TNEP validation, outside this PR.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -90,8 +102,90 @@ private Filing generateAccountFiling() {
...
+    }
+
+    /**
+     * Generates the filing based on the Filing model.
+     *
+     * @param transaction - transaction information
+     * @param accountsType - Account type information: account type, ixbrl's template name, account
+     * @param ixbrlLocation - the location where the ixbrl is stored.
+     * @return
+     * @throws IOException
```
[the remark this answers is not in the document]
ch-mcabrera: This used to throw an error before moving the call to the **Document Render Service** from here to the **Document generator**.  

I will remove it

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -90,8 +102,90 @@ private Filing generateAccountFiling() {
...
+        //TODO: call the document generator's new end point to get generate ixbrl. Implementation does NOT exist yet. (STORY SFA-595)
         return documentGeneratorCaller.generateIxbrl();
     }
 
+    /**
+     * Validates the ixbrl against TNEP. This validation is driven by the env. variable and it can
+     * be disable.
+     */
+    private boolean isValidIxbrl() {
+        boolean isIxbrlValid = true;
```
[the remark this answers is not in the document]
ch-mcabrera: This is a small method so it is easy to see there is only one exit point. Besides, this method could change when plugging in the TNEP validation. 
I will leave it as it is for now.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/FilingServiceImpl.java
```diff
@@ -90,8 +100,91 @@ private Filing generateAccountFiling() {
...
+    }
+
+    /**
+     * Get the description values, which currently contains the period end date. This data is
+     * required for the filing description.
+     *
+     * @param periodEndDate - account's period end date
+     * @return {@link Map<String,String>} containing period end date, to match filing model
+     */
+    private Map<String, String> getDescriptionValues(LocalDate periodEndDate) {
```
Mathsh: Just to make you aware, the new document-generator will be returning the description values in the api response. So not sure you will need this anymore?

ch-mcabrera: I've done this change since the Document generator implementation has not been completed yet.  I will wait until the Document generator's implementation is completed before removing this method.  
There is already a story to plug in the document generator and the story number has been added to a "TODO" where  document-generator's information will be used. I will updated the "TODO"

src/main/java/uk/gov/companieshouse/api/accounts/controller/FilingController.java
```diff
@@ -39,15 +39,15 @@ public ResponseEntity generateFiling(@PathParam("transactionId") String transact
 
         if (transaction == null) {
             logRequestError(request, "no transaction in request session");
-            return new ResponseEntity<>(HttpStatus.INTERNAL_SERVER_ERROR);
+            return new ResponseEntity<>(HttpStatus.BAD_REQUEST);
```
drosser: We can assume that the data is in a valid state (due to the interceptors) by the time it hits this controller, so if any of these objects are returning null, this should be an `INTERNAL_SERVER_ERROR` rather than a `BAD_REQUEST`. This applies to all of the changes you made regarding `BAD_REQUEST`.

PR #78, opened 2018-09-13
Add api validation for fixed assets

leahtaylor:
Add api validation for Balance Sheet, Fixed Assets and Shared-Capital Not Paid fields
Checks for mandatory fields, correct input range and correct totalling of fields.

Add validation annotations to CurrentPeriod, BalanceSheet and FixedAssets models with custom messages (@NotNull, @Range

Add BindingResultErrorToErrorMapping class to map binding result errors from annotations to our **Error** model, and collate these Errors into our **Errors** model

Add a **Current period validator** to validate totalling fields in FixedAssets
Add a **base validator** with reusable validation methods for other current period fields

Add **Error** and **Errors** models for the errors returned from validation

Add validation functionality to the create current period controller

Add enum classes
**ErrorMessagekeys** - For error messages returned from validation
**ErrorType** - for the Validation error types (Service or validation)
**LocationType** - for the error message location type (json-path)

SFA-327


leahtaylor:
screen shot 2018-09-12 at 15 30 08

leahtaylor:
screen shot 2018-09-14 at 10 40 16

leahtaylor:
screen shot 2018-09-14 at 10 58 44

src/main/java/uk/gov/companieshouse/api/accounts/utility/BindingResultErrorToErrorMapper.java
```diff
@@ -0,0 +1,65 @@
...
+                FieldError fieldError = (FieldError) object;
+
+                String field = fieldError.getField();
+                String errorMessage = fieldError.getDefaultMessage();
+
+                location = (field.replaceAll("(.)([A-Z])", "$1_$2")).toLowerCase();
+
+                if (errorMessage.equals("VALUE_OUTSIDE_RANGE")) {
+
+                    ArrayList arguments = new ArrayList();
```
sroberts2: I don't think this array is used after its created and has values assigned. `ArrayList arguments = new ArrayList();`

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    private void validateTotalFixedAssets(@Valid CurrentPeriod currentPeriod,
+        Errors errors) {
+        FixedAssets fixedAssets = currentPeriod.getBalanceSheet().getFixedAssets();
+        if (fixedAssets != null) {
+
+            Integer tangible = fixedAssets.getTangible();
+            Integer fixedAssetsTotal = fixedAssets.getTotalFixedAssets();
+            Integer calculatedTotal = tangible;
```
sroberts2: Why is tangible assigned to another Integer rather than used directly below as:
`validateAggregateTotal(fixedAssetsTotal, tangible, TOTAL_PATH, errors);`

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/BalanceSheet.java
```diff
@@ -1,14 +1,22 @@
 package uk.gov.companieshouse.api.accounts.model.rest;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.hibernate.validator.constraints.Range;
+
+import javax.validation.Valid;
+import javax.validation.constraints.Digits;
 import javax.validation.constraints.NotNull;
```
sroberts2: Unused imports

src/main/java/uk/gov/companieshouse/api/accounts/utility/BindingResultErrorToErrorMapper.java
```diff
@@ -0,0 +1,65 @@
...
+
+            if (object instanceof FieldError) {
+                FieldError fieldError = (FieldError) object;
+
+                String field = fieldError.getField();
+                String errorMessage = fieldError.getDefaultMessage();
+
+                location = (field.replaceAll("(.)([A-Z])", "$1_$2")).toLowerCase();
+
+                if (errorMessage.equals("VALUE_OUTSIDE_RANGE")) {
```
sroberts2: Can we write this line as `if ("VALUE_OUTSIDE_RANGE".equals(errorMessage)) {` to get a null check for free.

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Error.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Map;
+
+/**
+ * Represents a validation error
+ */
+public class Error {
+
+    @JsonProperty("error")
+    @Field("error")
+    private String error;
```
sroberts2: Sonar has raised an issue not sure how we tackle this given its copied from the spec:
> A field should not duplicate the name of its containing class

src/main/java/uk/gov/companieshouse/api/accounts/utility/BindingResultErrorToErrorMapper.java
```diff
@@ -0,0 +1,65 @@
...
+import uk.gov.companieshouse.api.accounts.validation.ErrorType;
+import uk.gov.companieshouse.api.accounts.validation.LocationType;
+
+import java.util.ArrayList;
+
+
+@Component
+public class BindingResultErrorToErrorMapper {
+
+    String location;
```
jterrych: This is a singleton component and cannot have state, this location should only be created locally.

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Error.java
```diff
@@ -0,0 +1,133 @@
...
+
+        this.error = error;
+        this.location = location;
+        this.locationType = locationType;
+        this.type = type;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
sroberts2: If we are going to auto-generate `equals` and `hashCode` methods can we use the Java standard template it is a lot easier to read and will be more closely aligned to our coding standards.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
leahtaylor: Because eventually the calculated total will contain the sum of all of the fixed assets fields which will be calculated here. Happy to move it until more fields are added?

src/main/java/uk/gov/companieshouse/api/accounts/utility/BindingResultErrorToErrorMapper.java
leahtaylor: This array list hold the parameters used for the @range validation (upper and lower limit)  this is then added to the error model when returned 

   ArrayList arguments = new ArrayList();
                    Object[] argument = fieldError.getArguments();
                    for (Object o : argument) {
                        arguments.add(o);
                    }
                    Error error = new Error(ErrorMessageKeys.VALUE_OUTSIDE_RANGE.getKey(),
                        location,
                        LocationType.JSON_PATH.getValue(), ErrorType.VALIDATION.getType());
                 ->   error.addErrorValue("lower", argument[2].toString());
                 ->   error.addErrorValue("upper", argument[1].toString());

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Error.java
leahtaylor: @jsimons1989 Should we be changing the specs to stop Sonar issues?

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Error.java
leahtaylor: This is the same method as used in the abridged project

src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
```diff
@@ -103,4 +138,22 @@ public ResponseEntity get(HttpServletRequest request) {
         return apiResponseMapper.mapGetResponse(responseObject.getData(), request);
 
     }
+
+    /**
+     * Log {@link AbridgedAccount} validation failure
```
sroberts2: AbridgedAccount?, It's a private method no need to add a Javadoc so you can just delete it.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
```diff
@@ -0,0 +1,45 @@
...
+public class CurrentPeriodValidator extends BaseValidator {
+
+    String CURRENT_PERIOD_PATH = "$.current_period";
+    String BALANCE_SHEET_PATH = CURRENT_PERIOD_PATH + ".balance_sheet";
+    String FIXED_ASSETS_PATH = BALANCE_SHEET_PATH + ".fixed_assets";
+    String TOTAL_PATH = FIXED_ASSETS_PATH + ".total";
+
+    public Errors validateCurrentPeriod(@Valid CurrentPeriod currentPeriod, Errors errors) {
+
+        if (currentPeriod.getBalanceSheet() != null) {
```
sroberts2: Null check validation occurs via the `@NotNull(message="MANDATORY_ELEMENT_MISSING")` annotation on the CurrentPeriod class no need to repeat yourself here.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
```diff
@@ -0,0 +1,45 @@
...
+
+            validateTotalFixedAssets(currentPeriod, errors);
+        }
+        return errors;
+    }
+
+    private void validateTotalFixedAssets(@Valid CurrentPeriod currentPeriod,
+        Errors errors) {
+        FixedAssets fixedAssets = currentPeriod.getBalanceSheet().getFixedAssets();
+        if (fixedAssets != null) {
```
sroberts2: Null check validation should be carried out via `@NotNull(message="MANDATORY_ELEMENT_MISSING")` annotation on the BalanceSheet class.

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Error.java
sroberts2: This will remain in place pending a further discussion with the architects it can be addressed as part of another change if needed.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
```diff
@@ -0,0 +1,45 @@
+package uk.gov.companieshouse.api.accounts.validation;
```
sroberts2: Should probably have an interface for this class. Possibly generic so `public interface Validator<C extends RestObject>` with a single method called `Errors validate(@Valid C rest, Errors errors);` not 100% sure though but certainly warrants a discussion.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
leahtaylor: Fixed assets is an optional field within the balance sheet so could be null even if the mandatory balance sheet isn't null - so I believe this check is required.

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Error.java
jsimons1989: @leahtaylor @sroberts2 The class can be rename to ValidationError to avoid the issue in Sonar.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
leahtaylor: It has been suggested we leave it as it is for now. We can always implement an interface at a later date if it is needed.

PR #79, opened 2018-09-17
Split abstract class to remove add link from child resources

sroberts2:
The child resources do not need the add link exposed or implemented, created an interface to model this.

Resolves: SFA-703


src/main/java/uk/gov/companieshouse/api/accounts/service/AbstractParentService.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.companieshouse.api.accounts.service;
+
+import uk.gov.companieshouse.api.accounts.LinkType;
+import uk.gov.companieshouse.api.accounts.exception.DataException;
+import uk.gov.companieshouse.api.accounts.model.rest.RestObject;
+
+public interface AbstractParentService<T extends RestObject> extends AbstractService<T> {
```
jsimons1989: The class name "AbstractParentService" suggests that this is an abstract class. I think the interface should be named what it is, a ParentService with a JavaDoc explaining what the interface is a contract for (it should be clear what the difference is between a service and a parent service without having to go looking at the code).

src/main/java/uk/gov/companieshouse/api/accounts/service/SmallFullService.java
```diff
@@ -4,6 +4,6 @@
 import uk.gov.companieshouse.api.accounts.model.rest.SmallFull;
 
 @Service
-public interface SmallFullService extends AbstractService<SmallFull> {
+public interface SmallFullService extends AbstractParentService<SmallFull> {
```
jsimons1989: Why is the CompanyAccountServiceImpl not also implementing the AbstractParentService? It also has an addLinks() method.

sroberts2: The CompanyAccount is a special case which could be made to fit the interface but doesn't fit the mould at the moment.

src/main/java/uk/gov/companieshouse/api/accounts/service/AbstractParentService.java
sroberts2: Agreed I will rename them.

PR #81, opened 2018-09-18
Add null check for balance sheet object 

leahtaylor:
Add null check for balance sheet object before validating fixed assets total to avoid null pointer


src/main/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidator.java
```diff
@@ -20,11 +20,15 @@
...
     public Errors validateCurrentPeriod(@Valid CurrentPeriod currentPeriod, Errors errors) {
+        
+        if(currentPeriod.getBalanceSheet()!=null) {
 
             validateTotalFixedAssets(currentPeriod, errors);
+        }
 
         return errors;
     }
+
```
caismith-ch: This extra line can be removed for consistency.

PR #84, opened 2018-09-20
Ensure 'period' is included in json path for returned errors

leahtaylor:
Ensure the full json path (including period) is included in binding result error json path.

Get period from 'Field error object' and append to beginning of the field error locations
Format return location to json format.

SFA-327


leahtaylor:
screen shot 2018-09-20 at 11 08 13

PR #88, opened 2018-09-24
Added error handling for json parse exceptions

sroberts2:
- Changed the `GlobalExceptionHandler` to extend `ResponseEntityExceptionHandler` so that we can return a `ResponseEntity` in every use case.
- Added language agnostic error messages for parse exceptions.
- Correct a Class name to use match Java format.
- Removed an unused import.
- Changed the error logging for `NoHandlerFoundException` and `MessageNotReadableException` from `ERROR` to `INFO` as this is user error.
- Added `json-body` as a location type.

Helps resolve: SFA-705

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
@@ -2,15 +2,12 @@ package uk.gov.companieshouse.api.accounts.exception.handler;
 
 import static uk.gov.companieshouse.api.accounts.CompanyAccountsApplication.APPLICATION_NAME_SPACE;
 
-import com.fasterxml.jackson.databind.exc.InvalidFormatException;
 import java.util.HashMap;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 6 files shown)


ChrisJBurns:
PR title? Also do we have a story number this resolves?

src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
```diff
@@ -2,42 +2,84 @@
...
+        return super.handleNoHandlerFoundException(ex, headers, status, request);
+    }
+
+    @Override
+    protected ResponseEntity<Object> handleHttpMessageNotReadable(
+        HttpMessageNotReadableException ex, HttpHeaders headers, HttpStatus status,
+        WebRequest request) {
+        logClientError(ex);
+
+        StringBuilder message = new StringBuilder("JSON parse exception");
```
ChrisJBurns: sorry to be that guy, but is there another way around the potential if else continuum?

src/main/java/uk/gov/companieshouse/api/accounts/validation/LocationType.java
```diff
@@ -5,6 +5,7 @@
  */
 public enum LocationType {
 
+    JSON_BODY("json-body"),
```
macreman: I think `request-body` would be better here to keep the spec addition more generic - also, can you add to the spec so we don't forget about it :)

PR #89, opened 2018-09-25
CompanyAccount service refactor to return Rest object.

sroberts2:
Refactored the Company account service implementation to return a `ResponseObject<CompanyAccount>` where applicable. 
Fixed the relevant calls to the Company account service.
Fixed the issues where we were using the ID from the `CompanyAccountEntity` model to use the CompanyID passed through in the URL.
Fixed the relevant Unit tests.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
@@ -1,22 +1,26 @@
 package uk.gov.companieshouse.api.accounts.controller;
 
+import static uk.gov.companieshouse.api.accounts.CompanyAccountsApplication.APPLICATION_NAME_SPACE;
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.servlet.http.HttpServletRequest;
+import javax.validation.Valid;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import org.springframework.validation.BindingResult;
-
+import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
```diff
@@ -66,23 +63,23 @@ public ResponseEntity create(@RequestBody @Valid CurrentPeriod currentPeriod,
 
         }
 
+        Map<String, String> pathVariables = (Map) request
```
macreman: if we can get this from the URI by adding a parameter to this controller method that would be nicer than using path variables

src/main/java/uk/gov/companieshouse/api/accounts/controller/CurrentPeriodController.java
```diff
@@ -111,16 +108,11 @@ public ResponseEntity get(HttpServletRequest request) {
...
             .getAttribute(AttributeName.TRANSACTION.getValue());
-        CompanyAccountEntity companyAccountEntity = (CompanyAccountEntity) request
-            .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
 
-        if (companyAccountEntity == null) {
-            LOGGER.error("Current Period error: No company account in request session",
-                logContext);
-            return ResponseEntity.status(HttpStatus.NOT_FOUND).body(null);
-        }
+        Map<String, String> pathVariables = (Map) request
```
macreman: same as above - if we can use a variable passed into the controller method that'd be nicer than pathvariables

src/main/java/uk/gov/companieshouse/api/accounts/controller/SmallFullController.java
```diff
@@ -44,22 +44,21 @@
...
-            HttpServletRequest request) {
+        HttpServletRequest request) {
         Transaction transaction = (Transaction) request
-                .getAttribute(AttributeName.TRANSACTION.getValue());
-        CompanyAccountEntity companyAccountEntity = (CompanyAccountEntity) request
-                .getAttribute(AttributeName.COMPANY_ACCOUNT.getValue());
-
-        String companyAccountId = companyAccountEntity.getId();
+            .getAttribute(AttributeName.TRANSACTION.getValue());
+        Map<String, String> pathVariables = (Map) request
```
macreman: as above - controller method param would be nicer

PR #96, opened 2018-09-27
Approval service layer classes and Unit test

sroberts2:
Added the `ApprovalService` class.
Added the `ApprovalServiceTest` class.

Correction for the Approvals repository and added the Resources name for "approval".

Helps Resolve: SFA-610

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalService.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalService.java
@@ -1,0 +1,125 @@
+package uk.gov.companieshouse.api.accounts.service.impl;
+
+import com.mongodb.MongoException;
+import java.util.HashMap;
+import java.util.Map;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DuplicateKeyException;
+import uk.gov.companieshouse.GenerateEtagUtil;
+import uk.gov.companieshouse.api.accounts.CompanyAccountsApplication;
+import uk.gov.companieshouse.api.accounts.Kind;
+import uk.gov.companieshouse.api.accounts.ResourceName;
+import uk.gov.companieshouse.api.accounts.exception.DataException;
+import uk.gov.companieshouse.api.accounts.links.BasicLinkType;
+import uk.gov.companieshouse.api.accounts.links.SmallFullLinkType;
+import uk.gov.companieshouse.api.accounts.links.TransactionLinkType;
+import uk.gov.companieshouse.api.accounts.model.entity.ApprovalEntity;
+import uk.gov.companieshouse.api.accounts.model.rest.Approval;
+import uk.gov.companieshouse.api.accounts.repository.ApprovalRepository;
+import uk.gov.companieshouse.api.accounts.service.ResourceService;
+import uk.gov.companieshouse.api.accounts.service.response.ResponseObject;
+import uk.gov.companieshouse.api.accounts.service.response.ResponseStatus;
+import uk.gov.companieshouse.api.accounts.transaction.Transaction;
+import uk.gov.companieshouse.api.accounts.transformer.ApprovalTransformer;
+import uk.gov.companieshouse.api.accounts.utility.impl.KeyIdGenerator;
+import uk.gov.companieshouse.logging.Logger;
+import uk.gov.companieshouse.logging.LoggerFactory;
+
+public class ApprovalService implements ResourceService<Approval> {
+
+    private static final Logger LOGGER = LoggerFactory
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalService.java
```diff
@@ -0,0 +1,125 @@
...
+        return new ResponseObject<>(ResponseStatus.FOUND, approval);
+    }
+
+    @Override
+    public String generateID(String companyAccountId) {
+        return keyIdGenerator.generate(companyAccountId + "-" + ResourceName.APPROVAL.getName());
+    }
+
+    public String createSelfLink(Transaction transaction, String companyAccountId) {
+        return transaction.getLinks().get(TransactionLinkType.SELF.getLink()) + "/"
```
ChrisJBurns: probably best to use a stringbuilder here?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @InjectMocks
+    private ApprovalService approvalService;
+
+    public void setUpCreate() {
+    }
+
+    @Test
+    @DisplayName("Tests the successful creation of an Approval resource")
+    public void canCreateAnApproval() throws DataException {
+        setUpCreate();
```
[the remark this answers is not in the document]
ChrisJBurns: I would also remove the method as well

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test
+    @DisplayName("Tests the mongo exception when creating an Approval")
+    void createApprovalMongoExceptionFailure() throws DataException {
+        setUpCreate();
+        doReturn(approvalEntity).when(approvalTransformer).transform(ArgumentMatchers
+            .any(Approval.class));
+        when(approvalRepository.insert(approvalEntity)).thenThrow(mongoException);
+        Executable executable = () -> {
+            approvalService.create(approval, transaction, "", "");
```
ChrisJBurns: Don't need to assign the executable here, can just do `assertThrows(DataException.class, approvalService.create(approval, transaction, "", ""));`

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Test
+    @DisplayName("Tests mongo exception thrown on find of an Approval resource")
+    public void findApprovalMongoException() throws DataException {
+        when(approvalRepository.findById("")).thenThrow(mongoException);
+        Executable executable = () -> {
+            approvalService.findById("", "");
+        };
+        assertThrows(DataException.class, executable);
```
ChrisJBurns: same here

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalService.java
sroberts2: It's easier to read as a String concatenation and is compiled to the same code regardless.

ChrisJBurns: Hmm I would have to disagree with the ease of reading as the reason I commented was because it was difficult to read what is being created. If not a `StringBuilder` then a `String.format`. But if you insist its easier to read then that's fair enough.

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
sroberts2: `assertThrows` only takes an Executable for the second parameter `approvalService.create(approval, transaction, "", "")` does not satisfy this?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
sroberts2: as above

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalService.java
sroberts2: IntelliJ disagrees I started to change the code to StringBuilder and it suggested I change it back to String to make it easier to read.

PR #97, opened 2018-09-28
Add previous fixed assets with validation to company-accounts-api

leahtaylor:
Add fixed assets to transformer

Add validation for previous-period fixed assets

- Add @Valid annotations to previous period model

- Add previous period validator to validate fixed assets totals

- Add validation to previous period controller

Add unit tests

SFA-448, SFA-449

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/PreviousPeriodController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/PreviousPeriodController.java
@@ -1,20 +1,23 @@
 package uk.gov.companieshouse.api.accounts.controller;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
-import org.springframework.web.bind.annotation.PostMapping;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RestController;
+import org.springframework.validation.BindingResult;
+import org.springframework.web.bind.annotation.*;
 import uk.gov.companieshouse.api.accounts.AttributeName;
 import uk.gov.companieshouse.api.accounts.exception.DataException;
 import uk.gov.companieshouse.api.accounts.model.entity.CompanyAccountEntity;
 import uk.gov.companieshouse.api.accounts.model.rest.PreviousPeriod;
+import uk.gov.companieshouse.api.accounts.model.validation.Errors;
 import uk.gov.companieshouse.api.accounts.service.impl.PreviousPeriodService;
 import uk.gov.companieshouse.api.accounts.service.response.ResponseObject;
 import uk.gov.companieshouse.api.accounts.transaction.Transaction;
 import uk.gov.companieshouse.api.accounts.utility.ApiResponseMapper;
+import uk.gov.companieshouse.api.accounts.utility.ErrorMapper;
+import uk.gov.companieshouse.api.accounts.validation.CurrentPeriodValidator;
+import uk.gov.companieshouse.api.accounts.validation.PreviousPeriodValidator;
 import uk.gov.companieshouse.logging.Logger;
 import uk.gov.companieshouse.logging.LoggerFactory;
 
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/controller/PreviousPeriodController.java
```diff
@@ -37,24 +44,45 @@
...
+
+        Errors errors = new Errors();
+
+        if (bindingResult.hasErrors()) {
+
+            errors = errorMapper.mapBindingResultErrorsToErrorModel(bindingResult, errors);
+
+        }
+
+        previousPeriodValidator.validatePreviousPeriod(previousPeriod, errors);
```
BSaunders95: Should we be performing validation if there are binding errors? This could open us up to NPE's if a non-null field were actually null.

src/main/java/uk/gov/companieshouse/api/accounts/validation/PreviousPeriodValidator.java
```diff
@@ -0,0 +1,43 @@
...
+        if (previousPeriod.getBalanceSheet() != null) {
+
+            validateTotalFixedAssets(previousPeriod, errors);
+        }
+
+        return errors;
+    }
+
+    private void validateTotalFixedAssets(@Valid PreviousPeriod previousPeriod,
+        uk.gov.companieshouse.api.accounts.model.validation.Errors errors) {
```
BSaunders95: Do we need the explicit class package here?

src/test/java/uk/gov/companieshouse/api/accounts/transformer/PreviousPeriodTransformerTest.java
```diff
@@ -10,14 +8,11 @@
...
 import uk.gov.companieshouse.api.accounts.model.rest.BalanceSheet;
-import uk.gov.companieshouse.api.accounts.model.rest.FixedAssets;
 import uk.gov.companieshouse.api.accounts.model.rest.PreviousPeriod;
 
 import java.util.HashMap;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.*;
```
BSaunders95: Can we make imports explicit please?

src/test/java/uk/gov/companieshouse/api/accounts/validation/PreviousPeriodValidatorTest.java
```diff
@@ -0,0 +1,44 @@
+package uk.gov.companieshouse.api.accounts.validation;
+
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import uk.gov.companieshouse.api.accounts.model.rest.BalanceSheet;
+import uk.gov.companieshouse.api.accounts.model.rest.CurrentPeriod;
+import uk.gov.companieshouse.api.accounts.model.rest.FixedAssets;
+import uk.gov.companieshouse.api.accounts.model.rest.PreviousPeriod;
+import uk.gov.companieshouse.api.accounts.model.validation.*;
```
BSaunders95: Same as above

src/test/java/uk/gov/companieshouse/api/accounts/validation/PreviousPeriodValidatorTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    String PREVIOUS_PERIOD_PATH = "$.previous_period";
+    String BALANCE_SHEET_PATH = PREVIOUS_PERIOD_PATH + ".balance_sheet";
+    String TOTAL_PATH = BALANCE_SHEET_PATH + ".fixed_assets.total";
+
+    PreviousPeriodValidator validator = new PreviousPeriodValidator();
+
+    PreviousPeriod previousPeriod = new PreviousPeriod();
+    BalanceSheet balanceSheet = new BalanceSheet();
+    uk.gov.companieshouse.api.accounts.model.validation.Errors errors = new uk.gov.companieshouse.api.accounts.model.validation.Errors();
```
BSaunders95: Again, do we need the explicit class package here?

src/main/java/uk/gov/companieshouse/api/accounts/controller/PreviousPeriodController.java
leahtaylor: The validator just validates the total field calculations, and performs null checks in the method.

leahtaylor:
screen shot 2018-10-01 at 06 57 49
screen shot 2018-10-01 at 06 57 22

leahtaylor:
screen shot 2018-10-01 at 06 59 16

src/main/java/uk/gov/companieshouse/api/accounts/validation/PreviousPeriodValidator.java
```diff
@@ -0,0 +1,42 @@
...
+import uk.gov.companieshouse.api.accounts.model.rest.FixedAssets;
+import uk.gov.companieshouse.api.accounts.model.rest.PreviousPeriod;
+import uk.gov.companieshouse.api.accounts.model.validation.Errors;
+
+import javax.validation.Valid;
+
+@Component
+public class PreviousPeriodValidator extends BaseValidator {
+
+    String PREVIOUS_PERIOD_PATH = "$.previous_period";
```
sroberts2: Can you add the accessors for these variables.

leahtaylor: do we need accessors for these? We don't have them in the current period validator and they are only used within this class

sroberts2: I presume you are declaring a constant given it is declared in all upper case, in which case we should be following our development standards in this regards. It should be prefixed with `private static final` or `public static final`. I suggest we change it in the CurrentPeriodValidator as well.

leahtaylor:
@sroberts2  have made the change so that if the binding result has errors, the custome validator is not run. This however means that we will never get all the errors together.

If an invalid character is passed, we will get a JSON parse error and no other errors
If a value out of range is passed (annotation validation) we will only get these and not the total errors
We will only get the total validation if there are no other errors

leahtaylor:
@sroberts2  @ChrisJBurns  'lenient; removed from unit tests and extra unit test added for when previous period validator fails in the controller

PR #102, opened 2018-10-09
Add validation for the approvals page

sroberts2:
Added a custom annotation to allow for Character set validation to be on a per field basis.
Added validation onto the Approvals model.

Resolves: SFA-611


src/main/java/uk/gov/companieshouse/api/accounts/validation/CharSetValid.java
```diff
@@ -0,0 +1,26 @@
...
+
+@Documented
+@Retention(RUNTIME)
+@Target(FIELD)
+@Constraint(validatedBy = CharSetValidatorImpl.class)
+public @interface CharSetValid {
+
+    String message() default "INVALID_VALUE";
+
+    CharSet value() default CharSet.CHARECTER_SET_1;
```
leahtaylor: Should this be 'CHARACTER' rather than CHARECTER ?

leahtaylor:
Looks good to me

PR #105, opened 2018-10-10
Add 'ifNotEmpty' checks around fields

leahtaylor:
Add 'ifNotEmpty' checks around fields  to display '0' values

SFA-752, SFA-453


BSaunders95:
Does `isNotEmpty` cater for nulls as well as empty strings?

leahtaylor:
Yes, it is a helper method we already have in the document render service that we used in abridged. https://github.com/companieshouse/document-render-service/blob/develop/helpers/strings.go. This seemed better to use than checking if each field was null, or 0..

PR #120, opened 2018-10-18
Added the validation for approvals

sroberts2:
Added a new class `ApprovalValidator`.
Added a new constraint annotation class `CharSetValid` and its implementation `CharSetValidatorImpl`
Implemented the above classes and tailored the `ResponseObject` and its associated classes to handle the this form of validation in the service layer.
Added/Amended the relevant Unit tests.

Resolves: SFA-611


src/main/java/uk/gov/companieshouse/api/accounts/validation/CharSetValidatorImpl.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.charset.CharSet;
+import uk.gov.companieshouse.charset.validation.CharSetValidation;
+
+@Component
+public class CharSetValidatorImpl implements ConstraintValidator<CharSetValid, String> {
+
+    @Autowired
+    CharSetValidation charSetValidation;
```
ChrisJBurns: is it worth making this privatte

src/test/java/uk/gov/companieshouse/api/accounts/utility/ApiResponseMapperTest.java
```diff
@@ -29,7 +29,7 @@
     RestObject restObject;
 
     @Mock
-    Errors validationErrorData;
+    Errors errors;
```
ChrisJBurns: worth making this private?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
```diff
@@ -102,6 +109,7 @@ public void createApprovalDuplicateKey() throws DataException {
     void createApprovalMongoExceptionFailure() throws DataException {
         doReturn(approvalEntity).when(approvalTransformer).transform(ArgumentMatchers
             .any(Approval.class));
+        doReturn(new Errors()).when(approvalValidator).validateApproval(approval, request);
         when(approvalRepository.insert(approvalEntity)).thenThrow(mongoException);
         Executable executable = () -> {
```
ChrisJBurns: i know this isn't in this PR but we could change bit of code to be one line?
FROM:
```java
Executable executable = () -> {
            approvalService.create(approval, transaction, "", request);
        };
        assertThrows(DataException.class, executable);
```

TO:
```java
        assertThrows(DataException.class, () -> approvalService.create(approval, transaction, "", request));
```

src/main/resources/ValidationMessages.properties
```diff
@@ -1,7 +1,11 @@
 #customise standard errors
 javax.validation.constraints.NotNull.message=mandatory_element_missing
+javax.validation.constraints.NotBlank.message=mandatory_element_missing
```
BSaunders95: There's a tab following this property which shouldn't be there and is causing issues when integrating with the web

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalService.java
```diff
@@ -45,27 +49,39 @@
...
+        final Map<String, Object> debugMap = new HashMap<>();
+        debugMap.put("transaction_id", transaction.getId());
+        debugMap.put("company_accounts_id", companyAccountId);
+        Errors errors = approvalValidator.validateApproval(rest, request);
+
+        if (errors.hasErrors()) {
+            DataException dataException = new DataException(
+                "Failed to validate " + ResourceName.APPROVAL.getName());
+            LOGGER.errorRequest(request, dataException, debugMap);
+            return new ResponseObject<>(ResponseStatus.VALIDATION_ERROR, rest, errors);
```
macreman: this shouldn't include the `rest` parameter as it will return an error instead of it so that argument should be null (or we could create a constructor that takes just the rest object and another that just takes the errors to avoid this going forwards

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/ApprovalServiceTest.java
sroberts2: The file is formatted to Companieshouse Java coding standards, I would be reluctant to change the code mentioned as I don't think a minor increase in readability warrants breaking the coding standards for. I could probably use a smaller naming convention for params to fit it on one line if you are really keen on this idea?

ChrisJBurns: It's not really a code standards issue, its more of a code efficiency issue, we are creating a redundant executable and then using it in the assert. Instead we can remove the assignment to an executable and just use the lambda in the assert. This saves the redundant assignment to the executable variable.

sroberts2: My mistake that change is made.

PR #126, opened 2018-10-25
Add models, transformer and repository for Accounting Policies

drosser:
- Add rest and entity models
- Add transformer (and tests)
- Add repository

Resolves: SFA-82.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/transformer/AccountingPoliciesTransformer.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/transformer/AccountingPoliciesTransformer.java
@@ -1,0 +1,34 @@
+package uk.gov.companieshouse.api.accounts.transformer;
+
+import org.springframework.beans.BeanUtils;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.accounts.model.entity.AccountingPoliciesDataEntity;
+import uk.gov.companieshouse.api.accounts.model.entity.AccountingPoliciesEntity;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/transformer/AccountingPoliciesTransformer.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.accounts.model.entity.AccountingPoliciesDataEntity;
+import uk.gov.companieshouse.api.accounts.model.entity.AccountingPoliciesEntity;
+import uk.gov.companieshouse.api.accounts.model.rest.AccountingPolicies;
+
+@Component
+public class AccountingPoliciesTransformer implements GenericTransformer<AccountingPolicies, AccountingPoliciesEntity> {
+
+    @Override
+    public AccountingPoliciesEntity transform(AccountingPolicies entity) {
```
JFisherCH: _minor comment_
Both transform methods take in a variable called "entity", is it worth calling this one something like policy instead? Just to make it a bit clearer that this one is taking in an AccountingPolicy and returning a AccountingPoliciesEntity and the other one is doing the reverse.

drosser: Will update it, the first one should arguably be called `rest` due to the difference between the rest/entity models

PR #131, opened 2018-10-31
Add StatementService to create, update and/or find the Legal Statements 

ch-mcabrera:
Changes to add the  Legal Statemt service. 
**Changes:** 
Add "statements" to the existing enums so that it can be used in the Statment service:
- kind
- ResourceName
- SmallFullLinkType

Add New properties , _LegalStatements.properties_ :
- This contains all the  Legal Statements that must be agreed to. By doing this, hardcode values are avoided in the Statement Service and it can be easily modified if it is needed (Removing existing or adding new statements).

Class added to load the _LegalStatements.properties_ :
- Sot that it can be used easier from the Statement Service.

Statement Service:
- Changes to add statement service and implement create, update and find.
- Logic around Legal Statements: **section_477**'s legal Statement contains an placeholder for the **period_end_on**, which is replaced by the **period_end_on** from the **Company Accounts**. 
This placeholder {period_end_on} could not be bound to any value as it gets updated in the service

Resolves: SFA-867

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementService.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementService.java
@@ -39,7 +39,7 @@ public class StatementService implements ResourceService<Statement> {
 
     private static final String PERIOD_END_ON_PLACE_HOLDER = "{period_end_on}";
     private static final String LEGAL_STATEMENT_SECTION_477_KEY = "section_477";
-    private static final DateTimeFormatter dateTimeFormatter = DateTimeFormatter
+    private static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormatter
         .ofPattern("d MMMM yyyy");
 
     private StatementTransformer transformer;
@@ -175,7 +175,6 @@ public class StatementService implements ResourceService<Statement> {
     private LocalDate getPeriodEndOn(CompanyAccount companyAccount) {
 
         return companyAccount.getPeriodEndOn();
-
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementService.java
```diff
@@ -0,0 +1,273 @@
...
+
+@Service
+public class StatementService implements ResourceService<Statement> {
+
+    private static final Logger LOGGER = LoggerFactory
+        .getLogger(CompanyAccountsApplication.APPLICATION_NAME_SPACE);
+
+    private static final String PERIOD_END_ON_PLACE_HOLDER = "{period_end_on}";
+    private static final String LEGAL_STATEMENT_SECTION_477_KEY = "section_477";
+    private static final DateTimeFormatter dateTimeFormatter = DateTimeFormatter
```
BSaunders95: The format of this variable should be capital cased, spaced with underscores. For example `DATE_TIME_FORMATTER`.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementService.java
```diff
@@ -0,0 +1,273 @@
...
+        return legalStatements;
+    }
+
+    /**
+     * Replace the place holder in the legal statement passed in, with the placeHolderReplacements
+     * value.
+     *
+     * @param statements
+     * @param legalStatementKey
+     * @param placeHolderReplacement
```
divya051988: @param placeHolder

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementService.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /**
+     * Sets the links, the etag, kind and statements(after replacing the placeholder) in the rest
+     * object.
+     *
+     * @param rest
+     * @param transaction
+     * @param companyAccountId
+     */
```
divya051988: @param periodEndOn

PR #136, opened 2018-11-05
Validate Director's name with size constraint

BSaunders95:
Validate Director's name provided on approval does not exceed 120 characters

Addresses bug raised in SFA-51


leahtaylor:
@BSaunders95 have you added this key to the message.properties in company-accounts.api?

BSaunders95:
@leahtaylor That change has already been made on the `feature/acc-policies-length-validation` branch 👍

PR #143, opened 2018-11-13
Fix Net current assets(liabilities) and Total less assets current liabilities showing as mandatory

caismith-ch:
Net current assets and Total less assets current liabilities are only mandatory when:
- CurrentAssets Total is not null
- Preprayments and accrued income is not null
- Creditors due within one year is not null

**Resolves**: SFA-904

### DIFF
--- a/src/test/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidatorTest.java
+++ b/src/test/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidatorTest.java
@@ -164,6 +164,39 @@ public class CurrentPeriodValidatorTest {
     }
 
     @Test
+    @DisplayName("ERROR - Mandatory elements missing for NetCurrentAssets and TotalAssetsLessCurrentLiabilities")
+    void validateMissingField() {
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/companieshouse/api/accounts/validation/CurrentPeriodValidatorTest.java
```diff
@@ -163,6 +163,39 @@ void validateBalanceSheetWithOtherLiabilitiesOrAssetsTotalNetAssetsError() {
                         ErrorType.VALIDATION.getType())));
     }
 
+    @Test
+    @DisplayName("ERROR - Mandatory elements missing for NetCurrentAssets and TotalAssetsLessCurrentLiabilities")
```
BSaunders95: Could we add tests for the scenario where either `CurrentAssets` or `OtherLiabilitiesOrAssets` is null while the other is not to improve unit test coverage?

BSaunders95:
Looks good barring one small comment on code coverage. The alternative way to perform this validation would be to add a custom class-level annotation which has the added benefit of validating the model before it even hits the controller.

PR #144, opened 2018-11-21
Refactor to trim all leading and trailing spaces from string input

sroberts2:
- Added the new class `DeserializerAdvice` which extends `SimpleModule` this registers the class as part of applications deserializers. The method contained is used for the String class only.

- Added a minimum value to some fields in the `AccountingPolicies` class and refactored the `ErrorMapper` class to handle returning this.

- Refactored the `@SpringBootApplication` class to remove the `System.exit(0);` call, the normal flow of the application should handle this scenario.

- Removed the creation of a custom `ObjectMapper` and added the relevant config to the initial Spring config for its internal `ObjectMapper` creation  - this will allow for easier manipulation of the ObjectMapper at a later date should we need it.

Resolves: SFA-939


src/main/java/uk/gov/companieshouse/api/accounts/DeserializerAdvice.java
```diff
@@ -0,0 +1,24 @@
...
+import com.fasterxml.jackson.databind.deser.std.StdScalarDeserializer;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import java.io.IOException;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DeserializerAdvice extends SimpleModule {
+
+    public DeserializerAdvice() {
+        addDeserializer(String.class, new StdScalarDeserializer<String>(String.class) {
```
rafahop: Just wondering if we could use `StringDeserializer`?

leahtaylor:
Couple of double line spaces but fine other than that

PR #148, opened 2018-11-27
Ensured a return for all error calls

sroberts2:
I've added a high level handler for exception to ensure we return http status 500 for an a failed call.

Resolves: SFA-931


src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
```diff
@@ -35,6 +37,16 @@
     @Value("${invalid.value}")
     public String invalidValue;
 
+    @ExceptionHandler(value = {Exception.class})
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
```
BSaunders95: Do we need this as well as the return? I believe you can make the method `void` or remove this annotation

src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
```diff
@@ -35,6 +37,16 @@
     @Value("${invalid.value}")
     public String invalidValue;
 
+    @ExceptionHandler(value = {Exception.class})
```
BSaunders95: No need for the explicit `value` or curly braces. The braces are used if you're capturing more than one type of exception. So this line can become `@ExceptionHandler(Exception.class)`

src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
```diff
@@ -35,6 +37,15 @@
     @Value("${invalid.value}")
     public String invalidValue;
 
+    @ExceptionHandler(Exception.class)
+    protected ResponseEntity<Object> handleException(Exception ex) {
+        HashMap<String, Object> message = new HashMap<>();
+        message.put("message", ex.getMessage());
+        message.put("error", ex.getClass());
+        STRUCTURED_LOGGER.error(ex.getMessage(), message);
```
BSaunders95: We're going to be logging the exception message twice; once explicitly and once as part of `message`. Is this required?

src/main/java/uk/gov/companieshouse/api/accounts/exception/handler/GlobalExceptionHandler.java
```diff
@@ -12,6 +12,8 @@
 import org.springframework.http.ResponseEntity;
 import org.springframework.http.converter.HttpMessageNotReadableException;
 import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.bind.annotation.ResponseStatus;
```
BSaunders95: I don't think this is used any more

PR #154, opened 2018-12-04
Add parent transaction link to company accounts resource

marcransome:
To support the [document generation changes being made](https://github.com/companieshouse/document-generator/pull/72) for small-full accounts, this change introduces a parent transaction link to the company accounts resource.


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyAccountServiceImpl.java
```diff
@@ -54,7 +56,7 @@
         String companyAccountLink = createSelfLink(transaction, id);
         companyAccount.setKind(Kind.COMPANY_ACCOUNTS.getValue());
         addEtag(companyAccount);
-        addLinks(companyAccount, companyAccountLink);
+        addLinks(companyAccount, companyAccountLink, transaction.getId());
```
jsimons1989: Could you use the transaction self link here instead of building the link?

PR #165, opened 2018-12-19
API - Add Delete to Debtors Note

caismith-ch:
Add the ability to delete a debtors note and return the correct response statuses.

**Resolves**: SFA-990

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DebtorsService.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DebtorsService.java
@@ -125,6 +125,11 @@ public class DebtorsService implements ResourceService<Debtors> {
     }
 
     @Override
+    public ResponseObject<Debtors> deleteById(String id, HttpServletRequest request) throws DataException {
+        return null;
+    }
+
+    @Override
     public String generateID(String companyAccountId) {
         return keyIdGenerator.generate(companyAccountId + "-" + ResourceName.DEBTORS.getName());
     }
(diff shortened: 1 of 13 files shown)


BSaunders95:
For the delete functionality, will we be removing the small full resource link?

caismith-ch:
Yeah I think you are right there, Will look into doing that.

src/test/java/uk/gov/companieshouse/api/accounts/controller/DebtorsControllerTest.java
```diff
@@ -242,9 +238,7 @@ void updateAccountingPoliciesSuccess() throws DataException {
     @DisplayName("Update debtors - data exception thrown")
     void updateAccountingPoliciesDataException() throws DataException {
 
-        when(mockRequest.getAttribute(anyString())).thenReturn(mockSmallFull).thenReturn(mockTransaction);
-        when(mockSmallFull.getLinks()).thenReturn(mockSmallFullLinks);
-        when(mockSmallFullLinks.get(SmallFullLinkType.DEBTORS_NOTE.getLink())).thenReturn("");
+        setupTransactionAndLinks();
```
ch-mcabrera: I would rename this to `mockTransactionAndLinks()`.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DebtorsService.java
```diff
@@ -134,6 +134,24 @@ public DebtorsService(DebtorsRepository repository,
         return new ResponseObject<>(ResponseStatus.FOUND, transformer.transform(entity));
     }
 
+    @Override
+    public ResponseObject<Debtors> deleteById(String id, HttpServletRequest request) throws DataException {
+        if (repository.existsById(id)) {
```
ch-mcabrera: I would move this statement to the `try and catch` as `repository.existsById(id)` throws a MongoException

```
try {
    if (repository.existsById(id)) {
        repository.deleteById(id);
        return new ResponseObject<>(ResponseStatus.UPDATED);
    } else {
        return new ResponseObject<>(ResponseStatus.NOT_FOUND);
    }

} catch (MongoException me) {
    ...
}
```

PR #175, opened 2019-01-09
Add Tangible Assets Models

BSaunders95:
Add tangible assets database entity and rest models and the `TangibleAssetsRepository` interface.

Required for SFA-935

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/entity/notes/tangible/DepreciationEntity.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/entity/notes/tangible/DepreciationEntity.java
@@ -1,0 +1,72 @@
+package uk.gov.companieshouse.api.accounts.model.entity.notes.tangible;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+
+public class DepreciationEntity {
+
+    @Field("at_period_end")
+    private Long atPeriodEnd;
+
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/model/entity/notes/tangible/DepreciationEntity.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Field("at_period_start")
+    private Long atPeriodStart;
+
+    @Field("charge_for_year")
+    private Long chargeForYear;
+
+    @Field("on_disposals")
+    private Long onDisposables;
+
```
leahtaylor: should both of these be disposables or disposals?

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/notes/tangible/Depreciation.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @JsonProperty("at_period_start")
+    private Long atPeriodStart;
+
+    @JsonProperty("charge_for_year")
+    private Long chargeForYear;
+
+    @JsonProperty("on_disposals")
+    private Long onDisposables;
+
```
leahtaylor: same here - should the son property match the field (disposals or disposables?)

PR #176, opened 2019-01-10
Return a new map from the statement service instead of the original

sroberts2:
Prevented a bug where this class from exposing the internal representation of the enclosed properties Map.

Resolves a bug on: SFA-829

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementsServiceProperties.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementsServiceProperties.java
@@ -16,4 +16,8 @@ public class StatementsServiceProperties {
     public Map<String, String> getCloneOfStatements() {
         return new HashMap<>(statements);
     }
+
+    public void setStatements(Map<String, String> statements) {
+        this.statements = statements;
+    }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/StatementsServiceProperties.java
```diff
@@ -13,7 +14,7 @@
     private Map<String, String> statements;
 
     public Map<String, String> getStatements() {
-        return statements;
+        return new HashMap<>(statements);
```
ljgeorge: Hi, is this different to initialising  **statements** like this:
**private Map<String, String> statements = new HashMap<>();**
and then just returning **statements** from the **getStatements()** method ?

sroberts2: It creates a new clone of the Map every time `getStatements` is called preventing the original Map from being altered. I could rename the method `getCopyOfStatements` to make it clearer what the intent is?

ljgeorge: I think your suggestion for the method name is good, would getCloneOfStatements ? be more accurate?

sroberts2: Agreed I will change it now.

PR #187, opened 2019-01-12
Add api validation for creditors within one year note

leahtaylor:
Add new validator for creditors within one year note
Rename model to be camel case creditorsWithinOneYear

Move isMultipleYearFiler method from debtors validator to company service to be reusable
Move add inconsistentData error to BaseValidator to be reusable

Add/update corresponding unit tests

SFA-385


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyServiceImpl.java
```diff
@@ -5,8 +5,10 @@
 import org.springframework.stereotype.Service;
 import org.springframework.web.util.UriTemplate;
 import uk.gov.companieshouse.api.ApiClient;
+import uk.gov.companieshouse.api.accounts.exception.DataException;
```
Joseph-Nash: This is an unused import in my IDE

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsWithinOneYearValidator.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void validatePreviousPeriod(@Valid CreditorsWithinOneYear creditorsWithinOneYear,
+            Errors errors) {
+        validatePreviousTotalCalculationCorrect(creditorsWithinOneYear, errors);
+        validatePreviousTotalProvidedIfValuesEntered(creditorsWithinOneYear, errors);
+    }
+
+    private void validatePreviousTotalProvidedIfValuesEntered(@Valid CreditorsWithinOneYear creditorsWithinOneYear, Errors errors) {
+        if ((creditorsWithinOneYear.getPreviousPeriod() != null) &&
```
JMetherell: Is it possible to reduce the number of creditorsWithinOneYear.getPreviousPeriod() and creditorsWithinOneYear.getCurrentPeriod() by using a local variable, or maybe just passing in the period rather than the full note into the individual validation methods?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyServiceImpl.java
```diff
@@ -41,5 +42,19 @@ public CompanyProfileApi getCompanyProfile(String companyNumber) throws ServiceE
...
+    @Override
+    public boolean isMultipleYearFiler (Transaction transaction) throws ServiceException {
+
+        try {
+            CompanyProfileApi companyProfile = getCompanyProfile(transaction.getCompanyNumber());
+            return (companyProfile != null && companyProfile.getAccounts() != null &&
+                    companyProfile.getAccounts().getLastAccounts() != null &&
+                    companyProfile.getAccounts().getLastAccounts().getPeriodStartOn() != null);
+
+        } catch (ServiceException e) {
```
BSaunders95: You don't need to catch this exception

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/notes/creditorsWithinOneYear/CreditorsWithinOneYear.java
```diff
@@ -1,4 +1,4 @@
-package uk.gov.companieshouse.api.accounts.model.rest.notes.creditorswithinoneyear;
+package uk.gov.companieshouse.api.accounts.model.rest.notes.creditorsWithinOneYear;
```
BSaunders95: Package name should follow convention and be all lower case

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsWithinOneYearValidator.java
```diff
@@ -0,0 +1,191 @@
...
+    @Autowired
+    public CreditorsWithinOneYearValidator(CompanyService companyService) {
+        this.companyService = companyService;
+    }
+
+    public Errors validateCreditorsWithinOneYear(@Valid CreditorsWithinOneYear creditorsWithinOneYear, Transaction transaction) throws DataException {
+
+        Errors errors = new Errors();
+
+        if (creditorsWithinOneYear != null) {
```
BSaunders95: We shouldn't ever need a null check on the resource. It _is_ better to program defensively, but the resource should exist by this point; when data is `POST`'ed or `PUT`, if Spring fails to map the JSON to an object and instead provides null we have bigger problems!

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsWithinOneYearValidator.java
```diff
@@ -0,0 +1,191 @@
...
+
+            addInconsistentDataError(errors,
+                    CREDITORS_WITHIN_PREVIOUS_PERIOD_TOTAL_PATH);
+        }
+    }
+
+    private void validatePreviousPeriod(@Valid PreviousPeriod creditorsPreviousPeriod,
+            Errors errors) {
+        validatePreviousTotalCalculationCorrect(creditorsPreviousPeriod, errors);
+        validatePreviousTotalProvidedIfValuesEntered(creditorsPreviousPeriod, errors);
```
BSaunders95: The logic in this method is ordered in the opposite way to what I would expect, personally.
Would it make more sense to...
1) check total is not null if other fields are provided
2) if errors are present, return - there isn't anything further to validate
3) else, validate the totals?

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsWithinOneYearValidator.java
```diff
@@ -0,0 +1,191 @@
...
+import uk.gov.companieshouse.api.accounts.exception.ServiceException;
+import uk.gov.companieshouse.api.accounts.model.rest.notes.creditorswithinoneyear.CreditorsWithinOneYear;
+import uk.gov.companieshouse.api.accounts.model.rest.notes.creditorswithinoneyear.CurrentPeriod;
+import uk.gov.companieshouse.api.accounts.model.rest.notes.creditorswithinoneyear.PreviousPeriod;
+import uk.gov.companieshouse.api.accounts.model.validation.Errors;
+import uk.gov.companieshouse.api.accounts.service.CompanyService;
+import uk.gov.companieshouse.api.accounts.transaction.Transaction;
+
+@Component
+public class CreditorsWithinOneYearValidator extends BaseValidator {
```
BSaunders95: Should we be cross validating against the balance sheet within this class?

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsWithinOneYearValidator.java
```diff
@@ -0,0 +1,191 @@
...
+                creditorsCurrentPeriod.getAccrualsAndDeferredIncome() != null ||
+                creditorsCurrentPeriod.getOtherCreditors() != null) &&
+                creditorsCurrentPeriod.getTotal() == null) {
+
+            addError(errors, invalidNote, CREDITORS_WITHIN_CURRENT_PERIOD_TOTAL_PATH);
+        }
+    }
+
+    private void validateDetailsPresentIfNoTotalProvided(@Valid CurrentPeriod creditorsCurrentPeriod, Errors errors) {
+        if (creditorsCurrentPeriod.getBankLoansAndOverdrafts() == null &&
```
BSaunders95: From the method name, I'd expect the logic to look like...
```
if (creditorsCurrentPeriod.getTotal() == null && creditorsCurrentPeriod.getDetails() == null) {
   // add error here
}
```
Previous methods should be sufficient to validate numeric fields.

Question is, should we ever accept a `null` total?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyServiceImplTests.java
```diff
@@ -68,16 +72,17 @@ void getCompanyProfileThrowsApiErrorResponseException() throws Exception {
...
 
         assertThrows(ServiceException.class, () ->
-            companyService.getCompanyProfile(COMPANY_NUMBER));
+                companyService.getCompanyProfile(COMPANY_NUMBER));
     }
 
     @Test
     @DisplayName("Get Company Profile - Throws URIValidationException")
-    void getBalanceSheetThrowsURIValidationException() throws ApiErrorResponseException, URIValidationException {
+    void getBalanceSheetThrowsURIValidationException() throws ApiErrorResponseException,
```
BSaunders95: Method name should be specific to company profile (I know this was pre-existing, reformatting just highlighted the problem)

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsWithinOneYearValidator.java
leahtaylor: We have a separate story for cross validation

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/CompanyServiceImplTests.java
```diff
@@ -38,6 +39,9 @@
     @Mock
     private CompanyGet mockCompanyGet;
 
+    @Mock
+    private Transaction mockTransaction;
```
BSaunders95: This is currently unused, I believe. Can we add unit test coverage for the `isMultipleYearFiler` method please?

pom.xml
```diff
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>uk.gov.companieshouse</groupId>
   <artifactId>company-accounts.api.ch.gov.uk</artifactId>
-  <version>unversioned</version>
+  <version>5cb2b632</version>
```
drosser: Should this be a hash or unversioned?

JMetherell: I didn't intend to commit that actually.  I'll set it back..

PR #191, opened 2019-01-15
Added tangible iXBRL template

drosser:
Resolves: SFA-356.


drosser:
> You will need to add formatting to your number fields for long figures to be displayed correctly (e.g. 999,999)
> 
> Use {{getMagnitudeFormatted $fieldName}}

Resolved

PR #192, opened 2019-01-17
Add field level validation for the tangible assets note

BSaunders95:
Add field level validation for the tangible assets note

Resolves SFA-353


src/main/java/uk/gov/companieshouse/api/accounts/model/rest/notes/tangible/TangibleAssetsResource.java
```diff
@@ -2,19 +2,28 @@
...
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import javax.validation.Valid;
+import org.hibernate.validator.constraints.Range;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class TangibleAssetsResource {
 
+    private static final int MAX_RANGE = 999999999;
+    private static final int MIN_RANGE = -999999999;
```
marcransome: Are the ranges defined here supposed to be one digit greater than those defined in `Depreciation.java` and `Cost.java`?

BSaunders95: Yes, net book value should max out at '9 nines'

PR #196, opened 2019-01-22
Add tangible assets to be part of balance sheet notes not additional notes

drosser:
It looks like there's lots of changes, but all I've done is moved the Tangible Assets so it's inside the 

`{{if $balanceSheetNotes := .small_full_accounts.balance_sheet_notes}}`

rather than inside 

`{{if $additionalNotes := .small_full_accounts.additional_notes}}`

and changed

`{{ if $tangibleAssetsNote := $additionalNotes.tangible_assets }}`

to be

`{{ if $tangibleAssetsNote := $balanceSheetNotes.tangible_assets }}`

### DIFF
--- a/src/main/resources/filing/templates/small-full-accounts.html
+++ b/src/main/resources/filing/templates/small-full-accounts.html
@@ -2589,432 +2589,432 @@ ul.accounts-comps-ct {
           {{end}}
           <!-- ACCOUNTING POLICIES END -->
 
-        {{ if $tangibleAssetsNote := $additionalNotes.tangible_assets }}
-            {{incValue "notes-page-count" }}
-            {{setValue "tangible-notes-id" (getValue "notes-page-count" ) }}
-            <div class="page">
-                <div class="header">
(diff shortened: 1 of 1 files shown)


src/main/resources/filing/templates/small-full-accounts.html
```diff
@@ -2589,432 +2589,434 @@ <h2>{{- getValue "notes-page-count"}}. Accounting policies </h2>
...
-            {{incValue "notes-page-count" }}
-            {{setValue "tangible-notes-id" (getValue "notes-page-count" ) }}
-            <div class="page">
-                <div class="header">
-                    <h1 class="text--center">
+        {{end}}
+        <!-- NOTES END -->
+
+        <!-- BALANCE SHEET NOTES START -->
+        {{if $balanceSheetNotes := .small_full_accounts.balance_sheet_notes}}
```
leahtaylor: Could we add similar comments for tangible note start and finish? Not essential, just makes for easier navigation in the template

PR #204, opened 2019-01-25
Add stocks controller, transformer, models and unit tests

marcransome:
These changes include the addition of:

* `StocksController` with `POST`, `PUT,` `GET` and `DELETE` handlers
* `StocksTransformer` with support for transforming API to web models and vice versa
* `StockService` with methods for manipulating the resource
* Unit tests for all three of the above

**Resolves:** `SFA-373`, `SFA-865`, `SFA-866`, `SFA-991`


src/main/java/uk/gov/companieshouse/api/accounts/controller/StocksController.java
```diff
@@ -0,0 +1,186 @@
...
+        Transaction transaction =
+                (Transaction) request.getAttribute(AttributeName.TRANSACTION.getValue());
+
+        ResponseEntity responseEntity;
+
+        try {
+            ResponseObject<Stocks> response = stocksService
+                    .create(stocks, transaction, companyAccountId, request);
+
+            responseEntity = apiResponseMapper
```
BSaunders95: You can just return here, no need for the variable assignment

src/main/java/uk/gov/companieshouse/api/accounts/controller/StocksController.java
```diff
@@ -0,0 +1,186 @@
...
+        SmallFull smallFull = (SmallFull) request.getAttribute(AttributeName.SMALLFULL.getValue());
+
+        Transaction transaction = (Transaction) request
+                .getAttribute(AttributeName.TRANSACTION.getValue());
+
+        String stocksId = stocksService.generateID(companyAccountsId);
+
+        try {
+            ResponseObject<Stocks> response = stocksService
+                    .deleteById(stocksId, request);
```
BSaunders95: You'll need to merge up from `develop`. The `ResourceService` interface now provides a `delete` method (formerly `deleteById`) which takes a `companyAccountsId`. This buys us the opportunity to call `SmallFullService#removeLink` within the delete method of the service to remove the resource link.

src/main/java/uk/gov/companieshouse/api/accounts/transformer/StocksTransformer.java
```diff
@@ -0,0 +1,69 @@
...
+        StocksEntity stocksEntity = new StocksEntity();
+
+        BeanUtils.copyProperties(rest, stocksDataEntity);
+        if (rest.getCurrentPeriod() != null) {
+            CurrentPeriodEntity currentPeriodEntity = new CurrentPeriodEntity();
+            BeanUtils.copyProperties(rest.getCurrentPeriod(), currentPeriodEntity);
+            stocksDataEntity.setCurrentPeriodEntity(currentPeriodEntity);
+        }
+
+        BeanUtils.copyProperties(rest, stocksDataEntity);
```
ch-mcabrera: Do we need two ` BeanUtils.copyProperties(rest, stocksDataEntity);`?

src/test/java/uk/gov/companieshouse/api/accounts/transformer/StocksTransformerTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        Stocks stocks = new Stocks();
+
+        stocks.setEtag(ETAG);
+        stocks.setKind(KIND);
+        stocks.setLinks(new HashMap<>());
+
+        CurrentPeriod currentPeriod = createCurrentPeriodRestObject();
+
+        stocks.setCurrentPeriod(currentPeriod);
```
ch-mcabrera: I would do the same here as you do in the other test `stocks.setCurrentPeriod(createCurrentPeriodRestObject());`

src/test/java/uk/gov/companieshouse/api/accounts/transformer/StocksTransformerTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+        CurrentPeriod currentPeriod = new CurrentPeriod();
+        currentPeriod.setPaymentsOnAccount(PAYMENTS_ON_ACCOUNT_CURRENT_PERIOD);
+        currentPeriod.setStocks(STOCKS_CURRENT_PERIOD);
+        currentPeriod.setTotal(TOTAL_CURRENT_PERIOD);
+
+        return currentPeriod;
+    }
+
+    private void assertEqualsEntityObject(StocksEntity stocksEntity) {
```
ch-mcabrera: we could call this something like `validateReturnedStockEntityObject()` instead of `assertEquals...`

ch-mcabrera: that's perfect. I like it more thant `assert....` ;)

PR #207, opened 2019-01-26
Implement cross validation for the tangible assets note

BSaunders95:
Implement cross validation for the tangible assets note

Resolves SFA-604


src/main/java/uk/gov/companieshouse/api/accounts/validation/TangibleAssetsValidator.java
```diff
@@ -910,6 +956,78 @@ private String getJsonPath(TangibleSubResource subResource, String pathSuffix) {
...
+            TangibleAssets tangibleAssets) throws DataException {
+
+        String currentPeriodId = currentPeriodService.generateID(companyAccountsId);
+
+        ResponseObject<CurrentPeriod> currentPeriodResponseObject =
+                currentPeriodService.findById(currentPeriodId, request);
+        CurrentPeriod currentPeriod = currentPeriodResponseObject.getData();
+
+        Long currentPeriodTangible =
+                Optional.ofNullable(currentPeriod)
```
sroberts2: We are using a combination of both lambda's and method references to do the same thing could we change it to use one or the other not both in the same file.

PR #209, opened 2019-01-29
Add stocks note to ixbrl template

JMetherell:
Resolves:
SFA-376


src/main/resources/filing/templates/small-full-accounts.html
```diff
@@ -2027,8 +2027,13 @@ <h2 class="text--center">Balance sheet</h2>
 
                       <tr class="figures">
                           <td id="stocks-label">Stocks:</td>
-                          <td class="figure noteIndex" id="uk-busTotalInventoriesCY_ENDNoteIndexes"></td>
-
+                          
+                        {{ if .small_full_accounts.balance_sheet_notes.stocks }}  
+                          <td class="figure noteIndex" id="uk-busTotalInventoriesCY_ENDNoteIndexes">{{getValue "stocks-notes-id"}}</td>
```
ch-mcabrera: I think this `id="uk-busTotalInventoriesCY_ENDNoteIndexes` should be the name instead of `id`
I was expecting to see an`ids` like:  `something-like-this-id`

src/main/resources/filing/templates/small-full-accounts.html
```diff
@@ -2155,8 +2160,12 @@ <h2 class="text--center">Balance sheet</h2>
 
                       <tr class="figures">
                           <td id="creditors-within-label">Creditors: amounts falling due within one year:</td>
-
-                          <td class="figure noteIndex" id="uk-busCreditorsCreditorsWithinOneYear_CY_ENDNoteIndexes"></td>
+                          
+                        {{ if .small_full_accounts.balance_sheet_notes.creditors_within_one_year }}
+                          <td class="figure noteIndex" id="uk-busCreditorsCreditorsWithinOneYear_CY_ENDNoteIndexes">{{getValue "creditors-within-one-year-notes-id"}}</td>
```
ch-mcabrera: Same here, name?

src/main/resources/filing/templates/small-full-accounts.html
```diff
@@ -2027,8 +2027,13 @@ <h2 class="text--center">Balance sheet</h2>
 
                       <tr class="figures">
                           <td id="stocks-label">Stocks:</td>
-                          <td class="figure noteIndex" id="uk-busTotalInventoriesCY_ENDNoteIndexes"></td>
-
+                          
+                          {{ if .small_full_accounts.balance_sheet_notes.stocks }}  
+                              <td class="figure noteIndex" id="uk-busTotalInventoriesCY_ENDNoteIndexes">{{getValue "stocks-notes-id"}}</td>
```
ch-mcabrera: I think this id="uk-busTotalInventoriesCY_ENDNoteIndexes should be the `name` instead of `id`
I was expecting to see anid like: `something-like-this-id`. i noticed there are other instances where this happens

ch-mcabrera: I spoke with @JMetherell and we are leaving it as it is since it matches the template. 
If this is wrong,  it will be fixed in a future PR

PR #212, opened 2019-02-04
Add stocks note validation, cross-validation and error handling

marcransome:
This pull-request introduces the following changes:

* Adds stocks note validation
* Adds stocks note cross-validation
* Adds stocks note error handling
* Adds validation tests and updates existing tests

**Resolves:** SFA-374, SFA-375, SFA-605

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/StocksValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/StocksValidator.java
@@ -1,0 +1,306 @@
+package uk.gov.companieshouse.api.accounts.validation;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import uk.gov.companieshouse.api.accounts.exception.DataException;
+import uk.gov.companieshouse.api.accounts.exception.ServiceException;
+import uk.gov.companieshouse.api.accounts.model.rest.BalanceSheet;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/StocksValidator.java
```diff
@@ -0,0 +1,306 @@
...
+                        stocks.getPreviousPeriod().getStocks() != null)) {
+
+            addError(errors, invalidNote, PREVIOUS_PERIOD_TOTAL_PATH);
+        }
+    }
+
+    private void validateCurrentPeriodTotalCorrect(Stocks stocks, Errors errors) {
+
+        if (stocks.getCurrentPeriod().getTotal() == null) {
+            return;
```
BSaunders95: Should we be returning here? What if `total` is null but `stocks` or `paymentsOnAccount` is not null?

marcransome: This method only validates that the total value provided is correct. We call `validateCurrentPeriodTotalPresent` before this method which should take care of instances where we have data for fields other than the `total`.

In the case where we have values for `stocks` or `payments_on_account` but no `total` an `invalid_note` error would be set (in `validateCurrentPeriodTotalPresent`) for the missing total and this method (`validateCurrentPeriodTotalCorrect`) would not add any additional errors (i.e. the total is either missing or incorrect and not both).

PR #215, opened 2019-02-05
Add employees note api

JMetherell:
Add entity model, controller, service and transformer

Resolves:
SFA-1078


src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
```diff
@@ -14,7 +14,8 @@
     INTANGIBLE_ASSETS_NOTE("small-full-accounts#intangible-assets-note"),
     SMALL_FULL_STATEMENT("small-full-accounts#statements"),
     STOCKS_NOTE("small-full-accounts#stocks-note"),
-    TANGIBLE_ASSETS_NOTE("small-full-accounts-note#tangible-assets");
+    TANGIBLE_ASSETS_NOTE("small-full-accounts-note#tangible-assets"),
+    EMPLOYEES_NOTE("small-full-accounts-note#employees-note");
```
BSaunders95: Do we need 'note' in both parts of the link? We should be consistent with the format of these. For reference, according to the spec the tangible link should be of the format: `small-full-accounts-note#tangible-assets` so I'd expect this to be `small-full-accounts-note#employees`

JMetherell: It really should be consistent - looks like stocks and intangible don't have note in the prefix either, but the spec says they should, so I'll update them all to match the spec - "small-full-accounts-note#foo"

src/main/java/uk/gov/companieshouse/api/accounts/transformer/EmployeesTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+
+        if (rest.getCurrentPeriod() != null) {
+            BeanUtils.copyProperties(rest, employeesDataEntity);
+            CurrentPeriodEntity currentPeriodEntity = new CurrentPeriodEntity();
+            BeanUtils.copyProperties(rest.getCurrentPeriod(), currentPeriodEntity);
+            employeesDataEntity.setCurrentPeriodEntity(currentPeriodEntity);
+        }
+
+        if (rest.getPreviousPeriod() != null) {
+            BeanUtils.copyProperties(rest, employeesDataEntity);
```
BSaunders95: Do we need this call in both 'if' blocks?

src/main/java/uk/gov/companieshouse/api/accounts/transformer/EmployeesTransformer.java
```diff
@@ -0,0 +1,68 @@
...
+        return employeesEntity;
+    }
+
+    @Override
+    public Employees transform(EmployeesEntity entity) {
+
+        Employees employees = new Employees();
+        EmployeesDataEntity employeesDataEntity;
+
+        if (entity.getData() != null) {
```
BSaunders95: Will this ever be `null`?

JMetherell: It wouldn't be in the normal course of events, but if the data in mongo isn't there as expected, the null check handles it gracefully.

src/test/java/uk/gov/companieshouse/api/accounts/transformer/EmployeesTransformerTest.java
```diff
@@ -0,0 +1,212 @@
...
+        CurrentPeriod currentPeriod = new CurrentPeriod();
+        currentPeriod.setAverageNumberOfEmployees(AVERAGE_NUMBER_OF_EMPLOYEES);
+        currentPeriod.setDetails(DETAILS);
+        return currentPeriod;
+    }
+
+
+    private PreviousPeriod createPreviousPeriodRestObject() {
+        PreviousPeriod previousPeriod = new PreviousPeriod();
+        previousPeriod.setAverageNumberOfEmployees(AVERAGE_NUMBER_OF_EMPLOYEES);
```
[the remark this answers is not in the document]
JMetherell: Updated as suggested

PR #217, opened 2019-02-06
Ensure correct errors being returned for first year filer (debtors)

marcransome:
These changes ensure that only a single `inconsistent_data` error is returned for the `previous_period` field when it has been supplied in a request for a first year filer.


marcransome:
Superseded by `feature/notes-validation-fixes`.

PR #221, opened 2019-02-11
Ensure consistent note validation error handling and messages

marcransome:
These changes ensure that existing notes return consistent errors that meet our expectations.


src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
```diff
@@ -60,10 +63,22 @@ protected void addIncorrectTotalError(Errors errors, String location) {
...
      * @param location
      */
-    protected void addInconsistentDataError (Errors errors, String location) {
+    protected void addInconsistentDataError(Errors errors, String location) {
 
         addError(errors, inconsistentData, location);
     }
+
+    /**
+     * Add an inconsistent data error for the given location
```
ch-mcabrera: Coudl we add more info to this method so that we can distinguish it from `protected void addInconsistentDataError (Errors errors, String location) {`

PR #228, opened 2019-02-19
Rename package and variables to comply with java standards

BSaunders95:
Required for SFA-1201


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DebtorsService.java
```diff
@@ -132,7 +132,7 @@ public DebtorsService (DebtorsRepository repository, DebtorsTransformer transfor
         } catch (MongoException e) {
             final Map<String, Object> debugMap = new HashMap<>();
             debugMap.put("id", id);
-            DataException dataException = new DataException("Failed to find Debtors", e);
+            DataException dataException = new DataException("Failed to find debtors", e);
```
drosser: Should this have changed?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DebtorsService.java
```diff
@@ -162,7 +162,7 @@ public DebtorsService (DebtorsRepository repository, DebtorsTransformer transfor
         } catch (MongoException me) {
             final Map<String, Object> debugMap = new HashMap<>();
             debugMap.put("id", companyAccountsId);
-            DataException dataException = new DataException("Failed to delete Debtors", me);
+            DataException dataException = new DataException("Failed to delete debtors", me);
```
drosser: Should this have changed?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DebtorsService.java
BSaunders95: No, it was an IDE update, good catch. Having looked at the class I've updated the message to use `ResourceName.DEBTORS.getName()` for consistency.

PR #237, opened 2019-02-26
Add employees api validation

leahtaylor:
Add validation for employees note api

Add @Range annotations to model classes
Add validation to check for correct periods on first year filer

Update logging and use of sdk for transactions in line with develop

Add unit tests

SFA-1083


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/EmployeesService.java
```diff
@@ -94,12 +109,8 @@ public EmployeesService(EmployeesRepository repository,
         try {
             repository.save(entity);
         } catch (MongoException me) {
-            DataException dataException =
-                    new DataException("Failed to update" + ResourceName.EMPLOYEES.getName(), me);
-            LOGGER.errorRequest(request, dataException, getDebugMap(transaction,
-                    companyAccountId, entity.getId()));
 
-            throw dataException;
+            throw new DataException(me);
```
BSaunders95: Can we remove the Logging from the `findById` method too please?

PR #238, opened 2019-02-26
Current assets investments note API 

Mathsh:
Adds POST/PUT/GET/DELETE, controller, service, transformer and models for the current assets investments note. Does not include validation.

Some ordering of imports picked up during changes. 

Resolves:
SFA-1135, 1136, 1137, 1138


src/main/java/uk/gov/companieshouse/api/accounts/links/SmallFullLinkType.java
```diff
@@ -14,7 +14,8 @@
     PREVIOUS_PERIOD("previous_period"),
     STATEMENTS("statements"),
     STOCKS_NOTE("stocks_note"),
-    TANGIBLE_ASSETS_NOTE("tangible_assets_note");
+    TANGIBLE_ASSETS_NOTE("tangible_assets_note"),
+    CURRENT_ASSETS_INVESTMENTS_NOTE("current-assets-investments-note");
```
leahtaylor: these should be underscores not hyphens

Mathsh: updated

PR #247, opened 2019-03-08
Throw empty_resource error if empty note is submitted

leahtaylor:
If creditors-after note resource and current/previous period is empty, throw 'empty-resource' error, otherwise continue with note validation.

Other notes to follow.

Add validateIfEmptyResource method to validator
Call new method from creditors-after service
Add addEmptyResourceError to base validator to use by other notes
Add new error to message.properties file
Update unit tests

SFA-1194

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
@@ -26,15 +26,15 @@ public class BaseValidator {
     protected String mandatoryElementMissing;
 
     @Value("${empty.resource}")
-    protected String emptyResource;
+    private String emptyResource;
 
     /**
      * Validate the given total is correctly aggregated
      *
-     * @param total
-     * @param expectedTotal
-     * @param location
-     * @param errors
+     * @param total actual total of the number fields
+     * @param expectedTotal expected total of the number fields
+     * @param location location json path location of the error
+     * @param errors errors errors object that holds any errors from submission
      */
     protected void validateAggregateTotal(Long total, Long expectedTotal, String location,
             Errors errors) {
@@ -50,9 +50,9 @@ public class BaseValidator {
     /**
      * Add an error for the given location
      *
-     * @param errors
-     * @param messageKey
-     * @param location
+     * @param errors errors errors object that holds any errors from submission
+     * @param messageKey relevent message key for the given error
+     * @param location location json path location of the error
      */
     protected void addError(Errors errors, String messageKey, String location) {
         errors.addError(new Error(messageKey, location, LocationType.JSON_PATH.getValue(),
@@ -62,8 +62,8 @@ public class BaseValidator {
     /**
      * Add an empty resource error for the given location
      *
-     * @param errors
-     * @param location
+     * @param errors errors object that holds any errors from submission
+     * @param location json path location of the error
      */
     public Errors addEmptyResourceError(Errors errors, String location) {
         addError(errors, emptyResource, location);
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
@@ -25,6 +25,9 @@ public class BaseValidator {
     @Value("${mandatory.element.missing}")
     protected String mandatoryElementMissing;
 
+    @Value("${empty.resource}")
+    protected String emptyResource;
+
     /**
      * Validate the given total is correctly aggregated
      *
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
```diff
@@ -25,6 +25,9 @@
     @Value("${mandatory.element.missing}")
     protected String mandatoryElementMissing;
 
+    @Value("${empty.resource}")
+    protected String emptyResource;
```
sroberts2: Can this be made private and use a getter and setter if required?

src/main/java/uk/gov/companieshouse/api/accounts/validation/BaseValidator.java
```diff
@@ -53,7 +56,19 @@ protected void validateAggregateTotal(Long total, Long expectedTotal, String loc
...
     protected void addError(Errors errors, String messageKey, String location) {
         errors.addError(new Error(messageKey, location, LocationType.JSON_PATH.getValue(),
-            ErrorType.VALIDATION.getType()));
+                ErrorType.VALIDATION.getType()));
+    }
+
+    /**
+     * Add an empty resource error for the given location
+     *
+     * @param errors
```
sroberts2: Add descriptions for @param or it will not render in a Javadoc.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsAfterOneYearValidator.java
```diff
@@ -70,26 +97,31 @@ public Errors validateCreditorsAfterOneYear(@Valid CreditorsAfterOneYear credito
...
+        } else if (validateCurrentPeriodExists(hasCurrentPeriodBalanceSheetNoteValue,
+                hasCurrentPeriodNoteData, errors)) {
 
             if (hasCurrentPeriodNoteData) {
                 validateCurrentPeriodFields(currentPeriodNote, errors);
             }
 
             if (hasCurrentPeriodNoteData && hasCurrentPeriodBalanceSheetNoteValue) {
-                crossValidateCurrentPeriodFields(currentPeriodNote, currentPeriodBalanceSheet, errors);
+                crossValidateCurrentPeriodFields(currentPeriodNote, currentPeriodBalanceSheet,
```
sroberts2: This if statement is always true.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsAfterOneYearValidator.java
```diff
@@ -106,26 +138,31 @@ private boolean validateNoUnexpectedDataPresent(boolean hasCurrentPeriodBalanceS
...
+        } else if (validatePreviousPeriodExists(hasPreviousPeriodBalanceSheetNoteValue,
+                hasPreviousPeriodNoteData, errors)) {
 
             if (hasPreviousPeriodNoteData) {
                 validatePreviousPeriodFields(previousPeriodNote, errors);
             }
 
             if (hasPreviousPeriodNoteData && hasPreviousPeriodBalanceSheetNoteValue) {
-                crossValidatePreviousPeriodFields(previousPeriodNote, previousPeriodBalanceSheet, errors);
+                crossValidatePreviousPeriodFields(previousPeriodNote, previousPeriodBalanceSheet,
```
sroberts2: This if statement is always true.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsAfterOneYearValidator.java
leahtaylor: It will be false if there is no balance sheet submitted, or no corresponding note value (eg creditors) is entered on the balance sheet and you try and submit the note (will only ever be false when submitting via api)

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsAfterOneYearValidator.java
leahtaylor: same as above

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsAfterOneYearValidator.java
sroberts2: The if statement on line 108 has `if (! hasCurrentPeriodBalanceSheetNoteValue &&` which ensures that `hasCurrentPeriodBalanceSheetNoteValue` is true by the time you reach line 122.

src/main/java/uk/gov/companieshouse/api/accounts/validation/CreditorsAfterOneYearValidator.java
sroberts2: same as above

PR #262, opened 2019-03-21
Change Find Method Signature

BSaunders95:
Refactor the resource service method signature 'findById' to 'find', taking a company accounts id variable, rather than a resource id. Resource id's are calculated within each service to delegate all business logic to the service layer.

Resolves SFA-1206

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/ResourceService.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/ResourceService.java
@@ -16,12 +16,10 @@ public interface ResourceService<T extends RestObject> {
         HttpServletRequest request)
         throws DataException;
 
-    ResponseObject<T> findById(String id, HttpServletRequest request)
+    ResponseObject<T> find(String companyAccountsId, HttpServletRequest request)
         throws DataException;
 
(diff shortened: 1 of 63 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/ResourceService.java
```diff
@@ -16,12 +16,10 @@
         HttpServletRequest request)
         throws DataException;
 
-    ResponseObject<T> findById(String id, HttpServletRequest request)
+    ResponseObject<T> find(String companyAccountsId, HttpServletRequest request)
```
drosser: If you're passing around `String`s representing `companyAccountsId`, you should be passing around the `CompanyAccounts` object instead and getting the ID where you need it

BSaunders95: The `CompanyAccount` REST object doesn't have an id exposed, what would be the suggested course of action?

drosser: We can't use the `CompanyAccount` REST object if it doesn't have a relevant ID on it, so I'll approve this as the `companyAccountsId` `String` is the right thing to use.

PR #265, opened 2019-04-03
Add CIC34 report models, transformer and service layer

BSaunders95:
Resolves SFA-1327


src/main/java/uk/gov/companieshouse/api/accounts/model/entity/Cic34ReportEntity.java
```diff
@@ -0,0 +1,26 @@
...
+        return data;
+    }
+
+    public void setData(Cic34ReportDataEntity data) {
+        this.data = data;
+    }
+
+    @Override
+    public String toString() {
+        return "Cic34ReportEntity {" +
```
Mathsh: can have toStrings as 

` @Override
    public String toString() {
        return new Gson().toJson(this);
    }`

as has been done in other places

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Cic34Report.java
```diff
@@ -0,0 +1,63 @@
...
+        return transferOfAssets;
+    }
+
+    public void setTransferOfAssets(String transferOfAssets) {
+        this.transferOfAssets = transferOfAssets;
+    }
+
+    @Override
+    public String toString() {
+        return "Cic34Report {" +
```
Mathsh: same as above regarding toString.

Mathsh:
Minor toString change, but everything else looks good

PR #271, opened 2019-04-09
Update the CIC34 report REST object to make all fields mandatory

BSaunders95:
Resolves SFA-1328


BSaunders95:
Closing due to spec changes

PR #293, opened 2019-05-24
Implement GET costs endpoint

BSaunders95:
Resolves SFA-1693


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/CostServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+        try {
+            List<PayableResource> payableResources = transactionService.getPayableResources(transaction);
+
+            if (payableResources.isEmpty()) {
+                return new Cost[0];
+            }
+
+            if (costs == null) {
+                costs =
```
sroberts2: This is config and should be read in at the start as opposed to on its first use, if its fails then we know theres an issue instantly.

PR #345, opened 2019-11-07
SFA:1937 Update the incorrect total validation for profit or loss before tax

divya051988:
- Total validation for profit or loss before tax.

Resolves : SFA:1937


src/main/java/uk/gov/companieshouse/api/accounts/validation/ProfitAndLossValidator.java
```diff
@@ -38,6 +39,7 @@ public Errors validateProfitLoss(@Valid ProfitAndLoss profitAndLoss, String comp
 
             validateGrossProfitTotal(profitAndLoss.getGrossProfitOrLoss(), errors);
             validateOperatingTotal(profitAndLoss, errors);
+            validateProfitOrLossBeforeTax(profitAndLoss,errors);
             if(errors.hasErrors()) {
```
faisal-amin82: I know you have not added these lines from (43 to 45). Do you think we really need it?

PR #382, opened 2020-01-14
Feature/balance sheet pl text fix

Jamie87:
Fixes a bug where the Statements page was displaying the wrong Profit and Loss statement regardless of whether a user was filing a P&L or not.
Resolves:
2042

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>uk.gov.companieshouse</groupId>
   <artifactId>company-accounts.api.ch.gov.uk</artifactId>
-  <version>0961512</version>
+  <version>unversioned</version>
   <packaging>jar</packaging>
   <name>company-accounts-api</name>
   <description>Company accounts API to handle company accounts data</description>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>uk.gov.companieshouse</groupId>
   <artifactId>company-accounts.api.ch.gov.uk</artifactId>
-  <version>unversioned</version>
+  <version>0961512</version>
   <packaging>jar</packaging>
   <name>company-accounts-api</name>
   <description>Company accounts API to handle company accounts data</description>
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
@@ -16,6 +16,7 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import uk.gov.companieshouse.api.accounts.AttributeName;
+import uk.gov.companieshouse.api.accounts.enumerator.AccountingPeriod;
 import uk.gov.companieshouse.api.accounts.exception.DataException;
 import uk.gov.companieshouse.api.accounts.links.CurrentPeriodLinkType;
 import uk.gov.companieshouse.api.accounts.links.PreviousPeriodLinkType;
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
@@ -79,12 +79,12 @@ public class ProfitAndLossController {
 
     @GetMapping
     public ResponseEntity get(@PathVariable("companyAccountId") String companyAccountId,
-                              HttpServletRequest request, AccountingPeriod period) {
+                              HttpServletRequest request, @PathVariable("accountingPeriod") AccountingPeriod accountingPeriod) {
 
         Transaction transaction = (Transaction) request.getAttribute(AttributeName.TRANSACTION.getValue());
 
         try {
-            ResponseObject<ProfitAndLoss> response = profitAndLossService.find(companyAccountId, period);
+            ResponseObject<ProfitAndLoss> response = profitAndLossService.find(companyAccountId, accountingPeriod);
             return apiResponseMapper.mapGetResponse(response.getData(), request);
         } catch (DataException ex) {
             LoggingHelper.logException(companyAccountId, transaction, "Failed to retrieve profit and loss resource",
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
@@ -16,7 +16,7 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import uk.gov.companieshouse.api.accounts.AttributeName;
-import uk.gov.companieshouse.api.accounts.enumerator.AccountingPeriod;
+import uk.gov.companieshouse.api.accounts.enumeration.AccountingPeriod;
 import uk.gov.companieshouse.api.accounts.exception.DataException;
 import uk.gov.companieshouse.api.accounts.links.CurrentPeriodLinkType;
 import uk.gov.companieshouse.api.accounts.links.PreviousPeriodLinkType;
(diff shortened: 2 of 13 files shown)


pom.xml
```diff
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>uk.gov.companieshouse</groupId>
   <artifactId>company-accounts.api.ch.gov.uk</artifactId>
-  <version>unversioned</version>
+  <version>0961512</version>
```
BSaunders95: Can you revert this please?

src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
```diff
@@ -53,7 +54,7 @@
     public ResponseEntity create(@Valid @RequestBody ProfitAndLoss profitAndLoss,
                                  BindingResult bindingResult,
                                  @PathVariable("companyAccountId") String companyAccountId,
-                                 HttpServletRequest request) {
+                                 HttpServletRequest request, AccountingPeriod period) {
```
BSaunders95: You'll need the `@PathVariable("accountingPeriod")` prefix

src/main/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossController.java
```diff
@@ -131,12 +132,12 @@ public ResponseEntity update(@RequestBody @Valid ProfitAndLoss profitAndLoss, Bi
 
     @DeleteMapping
     public ResponseEntity delete(@PathVariable("companyAccountId") String companyAccountId,
-                                 HttpServletRequest request) {
+                                 HttpServletRequest request, AccountingPeriod period) {
```
BSaunders95: And here

src/main/java/uk/gov/companieshouse/api/accounts/enumeration/AccountingPeriod.java
```diff
@@ -15,7 +15,7 @@ private String getPeriod() {
         return period;
     }
 
-    public static AccountingPeriod fromString(String period) {
+    public static AccountingPeriod  fromString(String period) {
```
BSaunders95: Careful on formatting - multiple spaces

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ProfitAndLossService.java
```diff
@@ -142,15 +147,17 @@ public ProfitAndLossService(
...
 
     @Override
-    public ResponseObject<ProfitAndLoss> delete(String companyAccountsId, HttpServletRequest request)
+    public ResponseObject<ProfitAndLoss> delete(String companyAccountsId, HttpServletRequest request, AccountingPeriod period )
         throws DataException {
 
-        String profitLossId = generateID(companyAccountsId, request);
+        String profitLossId = generateID(companyAccountsId, period);
+
+        invalidateStatementsIfExisting(companyAccountsId, request);
```
BSaunders95: Should this be further down? Perhaps just before `return new ResponseObject<>(ResponseStatus.UPDATED);` so we only invalidate statements if the db delete was successful?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ProfitAndLossService.java
```diff
@@ -195,15 +202,27 @@ private void initLinks(ProfitAndLoss profitAndLoss, String link) {
...
-    private boolean isCurrentPeriod(HttpServletRequest request) {
+    private boolean isCurrentPeriod(AccountingPeriod period) {
+
+        return period.equals(AccountingPeriod.CURRENT_PERIOD);
+    }
+
+    private void invalidateStatementsIfExisting(String companyAccountId, HttpServletRequest request)
+        throws DataException {
+
+        Statement statement = new Statement();
```
BSaunders95: Should this logic exist within the if? We may create a `Statement` object and never use it if the find returns a NOT_FOUND status. Same with the retrieval of the transaction from the request on the line below. We also need to set `hasAgreedToLegalStatements` to false on the `statement` object.

src/test/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossControllerTest.java
```diff
@@ -72,8 +73,8 @@
     @Mock
     private PreviousPeriod previousPeriod;
 
-    @Mock
-    private WebDataBinder webDataBinder;
+    //@Mock
```
BSaunders95: Please remove commented code

src/test/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossControllerTest.java
```diff
@@ -223,6 +236,8 @@ void updateProfitAndLossCurrentPeriodSuccess() throws DataException {
     @DisplayName("Tests the successful update of a profit and loss resource for previous period")
     void updateProfitAndLossPreviousPeriodSuccess() throws DataException {
 
+        AccountingPeriod period = AccountingPeriod.PREVIOUS_PERIOD;
```
BSaunders95: There's already an `accountingPeriod` variable declared below. Should we remove this and use that instead, or vice versa?

src/test/java/uk/gov/companieshouse/api/accounts/controller/ProfitAndLossControllerTest.java
```diff
@@ -336,6 +351,8 @@ void updateProfitAndLossPreviousPeriodBindingResultErrors() throws DataException
     @DisplayName("Tests the update of an profit and loss resource where the service throws a data exception")
     void updateProfitLossServiceThrowsDataException() throws DataException {
 
+        AccountingPeriod period = AccountingPeriod.CURRENT_PERIOD;
```
BSaunders95: As above

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/StatementServiceTest.java
```diff
@@ -52,14 +53,17 @@
 
     private static final String ETAG = "etag";
     private static final String LEGAL_STATEMENT_ID = "abcdef";
-    private static final String COMPANY_ACCOUNTS_ID = "123123";
+    private static final String COMPANY_ACCOUNTS_ID = "";
```
BSaunders95: Why has this changed? Can it be reverted back please?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/StatementServiceTest.java
```diff
@@ -115,12 +121,17 @@ void shouldCreateStatement() throws DataException {
         when(transactionMock.getLinks()).thenReturn(transactionLinksMock);
         when(transactionLinksMock.getSelf()).thenReturn(SELF_LINK);
 
+        ResponseObject responseObject = new ResponseObject(ResponseStatus.FOUND);
+        when(profitAndLossService.find(COMPANY_ACCOUNTS_ID, uk.gov.companieshouse.api.accounts.enumeration.AccountingPeriod.CURRENT_PERIOD)).
```
BSaunders95: Do we need the fully qualified package name here?

Jamie87: Yes because we already have an AccountingPeriod imported from model

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/StatementServiceTest.java
```diff
@@ -1,6 +1,7 @@
 package uk.gov.companieshouse.api.accounts.service.impl;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
```
BSaunders95: Can we add a test or two to this class for the case where `ProfitAndLossService#find` returns a `NOT_FOUND` status?

src/test/java/uk/gov/companieshouse/api/accounts/service/impl/StatementServiceTest.java
BSaunders95: 👍

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/ProfitAndLossService.java
```diff
@@ -195,15 +203,29 @@ private void initLinks(ProfitAndLoss profitAndLoss, String link) {
...
+    private boolean isCurrentPeriod(AccountingPeriod period) {
+
+        return period.equals(AccountingPeriod.CURRENT_PERIOD);
+    }
 
-        return CURRENT_PERIOD_PATTERN.matcher(request.getRequestURI()).matches();
+    private void invalidateStatementsIfExisting(String companyAccountId, HttpServletRequest request)
+        throws DataException {
+
+        if (statementService.find(companyAccountId, request).getStatus().equals(ResponseStatus.FOUND)) {
```
BSaunders95: The NOT_FOUND case is never covered in unit tests

PR #388, opened 2020-01-31
Add new context ref for accounts approval director

faisal-amin82:
- Add a new context ref to avoid the validation error on ixbrl template for the accounts director
- Currently we have two Director1_CY in our template which is a validation issue

Resolves: SFA-2452


perseus12:
This resolves bug SFA-2452
https://companieshouse.atlassian.net/browse/SFA-2452

PR #389, opened 2020-02-05
Approval api validation for name

faisal-amin82:
- Add code to resolve the approval name issue where directors report is present and approval name does not match the name of the serving directors included in the directors report

Resolves: SFA-2468

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/ApprovalValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/ApprovalValidator.java
@@ -66,7 +66,7 @@ public class ApprovalValidator extends BaseValidator {
 
             if (servingDirectors != null && !servingDirectors.isEmpty()) {
 
-                if (!(servingDirectors.stream().noneMatch(d -> !d.equalsIgnoreCase(approval.getName())))) {
+                if (servingDirectors.stream().noneMatch(d -> d.equalsIgnoreCase(approval.getName()))) {
                     errors.addError(new Error(invalidValue, NAME_PATH, LocationType.JSON_PATH.getValue(), ErrorType.VALIDATION.getType()));
                 }
             }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/ApprovalValidator.java
```diff
@@ -31,7 +49,29 @@ public Errors validateApproval(Approval approval, HttpServletRequest request) {
...
+                    .orElse(null);
+
+            List<String> servingDirectors = new ArrayList<>();
+
+            if (directors != null) {
+
+                Arrays.stream(directors).filter(d -> d.getResignationDate() == null).forEach(director -> servingDirectors.add(director.getName()));
+            }
+
+            if (servingDirectors != null && !servingDirectors.isEmpty()) {
```
BSaunders95: You don't need the null check here; you instantiated `servingDirectors` so it can never be null

faisal-amin82: Removed the extra null check that was not required

src/main/java/uk/gov/companieshouse/api/accounts/validation/ApprovalValidator.java
```diff
@@ -31,7 +49,25 @@ public Errors validateApproval(Approval approval, HttpServletRequest request) {
                 ErrorType.VALIDATION.getType()));
         }
 
+        ResponseObject<Director> directorsReportResponseObject = directorService.findAll(transaction, companyAccountId, request);
+
+        if(directorsReportResponseObject != null) {
```
BSaunders95: For future ref, you don't need this null check either; directorService#findAll will never return null

faisal-amin82: Removed the unwanted if and sorted the tests out

PR #407, opened 2020-02-28
Styling changes for cic report

faisal-amin82:
- Cic styling changes according to the prototype

Resolves: SFA-2426


BSaunders95:
Looks great, but we may need to add 'CIC report' to the front page (which admittedly wasn't part of the prototype)

PR #424, opened 2020-04-02
Resolve issues with stocks validator

faisal-amin82:
- Issues to resolve where karate tests fails for stocks validator

Resolves: SFA-2561

### DIFF
--- a/src/test/java/uk/gov/companieshouse/api/accounts/validation/StocksValidatorTest.java
+++ b/src/test/java/uk/gov/companieshouse/api/accounts/validation/StocksValidatorTest.java
@@ -97,36 +97,6 @@ public class StocksValidatorTest {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/companieshouse/api/accounts/validation/StocksValidatorTest.java
```diff
@@ -181,7 +183,9 @@ void testSuccessfulCrossValidationForMultipleYearFiler() throws DataException {
 
     @Test
     @DisplayName("Cross validation fails with valid note for multiple year filer")
-    void testErrorReturnedCrossValidationForMultipleYearFiler() throws DataException {
+    void testErrorReturnedCrossValidationForMultipleYearFiler() throws DataException, ServiceException {
```
BSaunders95, line 186: These tests are now duplicates; they used to test `crossValidate` which has now been totally removed from the class. Shall we bin them?

PR #427, opened 2020-04-06
Refactor debtors to use generic note controller

BSaunders95:
Refactor debtors to use generic note controller.

Resolves SFA-2566


src/main/java/uk/gov/companieshouse/api/accounts/repository/smallfull/DebtorsRepository.java
```diff
@@ -0,0 +1,15 @@
...
+import uk.gov.companieshouse.api.accounts.model.entity.smallfull.notes.debtors.DebtorsEntity;
+import uk.gov.companieshouse.api.accounts.repository.AccountsNoteRepository;
+
+@Repository
+public interface DebtorsRepository extends AccountsNoteRepository<DebtorsEntity> {
+
+    @Override
+    default AccountingNoteType getAccountsNote() {
+        return AccountingNoteType.SMALL_FULL_DEBTORS;
+    }
```
rafahop, line 14: I wonder if Spring would accept it if this was a method of an abstract class rather than a default method of an interface? It would make more sense that way if it worked

BSaunders95: I'll definitely look into it, cheers for the suggestion @rafahop

PR #429, opened 2020-04-07
Tidy pom file

rafahop:
Remove unnecessary properties, profiles, plugins and repositories

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,20 +30,10 @@
     <api-security-java.version>0.2.9</api-security-java.version>
     <commons-io.version>2.4</commons-io.version>
 
-    <!-- Maven and Surefire plugins -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -128,146 +118,45 @@
...
-            <dependencies>
-              <dependency>
-                <groupId>org.junit.platform</groupId>
-                <artifactId>junit-platform-surefire-provider</artifactId>
-                <version>${junit-platform-surefire-provider.version}</version>
-              </dependency>
-            </dependencies>
-          </plugin>
-          <plugin>
-            <groupId>org.jacoco</groupId>
```
BSaunders95, line 238: There were issues removing this before I believe. Sonar test coverage dropped to zero. So it may be worth checking analysis of this PR prior to merging just in case 👍

PR #438, opened 2020-04-29
Debtors additional information validation rework

amoriartyCH:
-Changed validation to allow users to progress if they only provide "additional information" details in the current period And leave the current period blank on the balance sheet.

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/DebtorsValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/DebtorsValidator.java
@@ -99,21 +99,13 @@ public class DebtorsValidator extends BaseValidator implements NoteValidator<Deb
             BalanceSheet currentPeriodBalanceSheet, Errors errors) {
 
         boolean hasCurrentPeriodBalanceSheet = currentPeriodBalanceSheet != null;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/DebtorsValidator.java
```diff
@@ -160,9 +152,9 @@ private boolean validateNoUnexpectedDataPresent(boolean hasCurrentPeriodBalanceS
     }
 
     private void validateCurrentPeriodFields(CurrentPeriod debtorsCurrentPeriod, Errors errors) {
-        if (debtorsCurrentPeriod.getTotal() == null) {
+        if (debtorsCurrentPeriod.getTotal() == null && debtorsCurrentPeriod.getDetails() == null) {
```
BSaunders95, line 155: Should we do a null or empty check, else it'll be possible to submit a current period with additional info set to `""`? Maybe use apache commons' `StringUtils#isBlank`?

amoriartyCH: As suggested. Changed to use the isBlank method to guard against white space and empty submissions. Thanks!

PR #450, opened 2020-06-12
Balance sheet statements ARD update

mreis-ch:
Add method to StatementService to set agreed legal statement to false if they exist
Add calls to above method from the update method in SmallFullService
Update calls to method in the create and delete methods in ProfitAndLossService

BI-3184

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/SmallFullService.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/SmallFullService.java
@@ -12,7 +12,6 @@ import org.springframework.stereotype.Service;
 import com.mongodb.MongoException;
 
 import uk.gov.companieshouse.GenerateEtagUtil;
-import uk.gov.companieshouse.api.accounts.AttributeName;
 import uk.gov.companieshouse.api.accounts.Kind;
 import uk.gov.companieshouse.api.accounts.ResourceName;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/SmallFullService.java
```diff
@@ -49,6 +49,9 @@
 
     private CompanyService companyService;
 
+    @Autowired
```
ch-tstrickland, line 52: Constructor is already autowired, which is how we should be doing it everywhere anyway, so this field ought to be not autowired AND be passed into the autowired constructor.

ch-tstrickland:
Perhaps you forgot to check in the new unit tests?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/SmallFullService.java
mreis-ch: I did initially put the StatementService in the constructor but on start up was getting a cyclic error which I couldn't quite sort. I will take another look

ch-tstrickland: I seem to remember Brad saying that this was a reason to do it this way - that it shows up this kind of issue.

PR #460, opened 2020-07-30
Add loan validation

mreis-ch:
Create new class LoanValidator with unit test
Update LoanServiceImpl to call LoanValidator

BI-4040

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/LoanValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/LoanValidator.java
@@ -1,0 +1,74 @@
+package uk.gov.companieshouse.api.accounts.validation;
+
+import org.springframework.stereotype.Component;
+
+import uk.gov.companieshouse.api.accounts.model.rest.smallfull.notes.loanstodirectors.Loan;
+import uk.gov.companieshouse.api.accounts.model.rest.smallfull.notes.loanstodirectors.LoanBreakdownResource;
+import uk.gov.companieshouse.api.accounts.model.validation.Errors;
+
+@Component
+public class LoanValidator extends BaseValidator {
+
+    private static final String LOANS_PATH = "$.loans";
+
+    private static final String LOANS_DIRECTOR_NAME = LOANS_PATH + ".director_name";
+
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/LoanValidator.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String LOANS_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START = LOANS_PATH + ".breakdown.balance_at_period_start";
+
+    private static final String LOANS_BREAKDOWN_PATH_BALANCE_AT_PERIOD_END = LOANS_PATH + ".breakdown.balance_at_period_end";
+
+
+	public Errors validateLoan(Loan loan) {
+
+		Errors errors = new Errors();
+
+		if (loan.getDirectorName() == null) {
```
BSaunders95, line 31: We could enforce these conditions using `@NotNull` annotations on the rest model fields - this would be more succinct, but disguises the logic a bit, so really a matter of preference

src/main/java/uk/gov/companieshouse/api/accounts/validation/LoanValidator.java
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		if (loan.getDescription() == null) {
+            addError(errors, mandatoryElementMissing, LOANS_DESCRIPTION);
+		}
+
+		LoanBreakdownResource loanBreakdown = loan.getBreakdown();
+		Boolean missingBreakdownElement = false;
+		
+		if (loanBreakdown.getAdvancesCreditsMade() == null) {
```
BSaunders95, line 42: If `loanBreakdown` is null, this will throw an NPE

mreis-ch: Yeah, picked up by Aaron. Made changes 👍

src/main/java/uk/gov/companieshouse/api/accounts/validation/LoanValidator.java
mreis-ch: Is there a preference on generally whether to go with annotations rather than logic?

PR #461, opened 2020-08-03
Add validation to Loan

mreis-ch:
Add valid character and length validation annotations to Loan.java
Add range validation annotations to LoanBreakdownResource.java
Add tests for validation

BI-4529

### DIFF
--- a/src/test/java/uk/gov/companieshouse/api/accounts/controller/LoansControllerTest.java
+++ b/src/test/java/uk/gov/companieshouse/api/accounts/controller/LoansControllerTest.java
@@ -25,15 +25,18 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/companieshouse/api/accounts/controller/LoansControllerTest.java
```diff
@@ -120,6 +134,21 @@ void createLoanAndServiceThrowsDataException() throws DataException {
         verify(apiResponseMapper, times(1))
                 .getErrorResponse();
     }
+    
+    @Test
+    @DisplayName("Tests the creation of a Loan where the controller returns a bad request binding error for invalid length")
+    void createLoanAndBadRequestResponseServiceThrowsDataException() throws DataException {
```
BSaunders95, line 140: Want to include a similar test for `update`?

mreis-ch: Ah, good point. Will add one 👍

PR #466, opened 2020-08-03
Implement validation for loans additional info

BSaunders95:
Resolves BI-4734, BI-4740

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/controller/LoansToDirectorsAdditionalInformationController.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/controller/LoansToDirectorsAdditionalInformationController.java
@@ -1,9 +1,11 @@
 package uk.gov.companieshouse.api.accounts.controller;
 
+import javax.validation.Valid;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/controller/LoansToDirectorsAdditionalInformationController.java
```diff
@@ -69,6 +83,11 @@ public ResponseEntity update(@RequestBody AdditionalInformation additionalInform
             return new ResponseEntity<>(HttpStatus.NOT_FOUND);
         }
 
+        if (bindingResult.hasErrors()) {
```
mreis-ch, line 86: Would you do this before looking up the LoansToDirectors links?

BSaunders95: I considered this, but I think regardless of whether data submitted is valid or not we ought to check a resource exists first. Imo, if a user were to PUT bad data against a non-existent resource, a `404` would trump a `400` but I'm happy to defer to the team if they have stronger feelings either way

PR #477, opened 2020-09-08
Loans to directors IXBRL template

BSaunders95:
Resolves BI-4408, BI-4478

### DIFF
--- a/src/main/resources/filing/templates/small-full-accounts.html
+++ b/src/main/resources/filing/templates/small-full-accounts.html
@@ -4431,9 +4431,11 @@
                 <td colspan="2">Description of the transaction:</td>
               </tr>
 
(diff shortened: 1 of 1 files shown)


ch-tstrickland (commented):
How do we go about testing this? Is it by eye-of-developer: bash it around until the output looks right, then get a tester to do the same?

BSaunders95:
> How do we go about testing this? Is it by eye-of-developer: bash it around until the output looks right, then get a tester to do the same?

Exactly that I'm afraid @ch-tstrickland - there aren't any testing tools we can use for this (save for a validator, which checks the underlying ixbrl entities)

ch-tstrickland (approved):
Approved on the understanding that Brad has basically coded this until it looks right & our testers will give a good kicking too.

PR #480, opened 2020-09-11
Autodelete Loans To Director loans on director change

ch-tstrickland:
When a director is created/updated/deleted, then the Loans To Director's resource's loans become uncertain. The cleanest, easiest solution is to delete them and prompt the user to re-enter the lot. This is unlikely to happen, as the user will usually have entered all the directors and then their loans, so this is safe.

Implement logic and unit tests.


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/DirectorService.java
```diff
@@ -231,4 +254,22 @@ private String getDirectorId(HttpServletRequest request) {
 
         return directorId;
     }
+
+    private void directorsResourceChanged(Transaction transaction, String companyAccountsId,
```
BSaunders95, line 258: All looks good, though I may recommend renaming this method to be more descriptive. Perhaps `removeAssociatedLoans` or similar?

BSaunders95 (approved):

PR #483, opened 2020-09-17
Add validation check for multi-year filer to LoanValidator

mreis-ch:
Add checks to LoanValidator to check if company is a multi year filer. If not then do not use the balance at period start in calculations.
Added method, getIsMultipleYearFiler, to BaseValidator. Copied from and then removed from EmployeesValidator. This method was duplicated in other validators and removed to use the one in BaseValidator. Knock on effect other validators as required changes to their constructors or constructors added.
Added getter and setter for CompanyService in BaseValidator and replaced called to the getter in other validators that used CompanyService
Updated tests accordingly to cater for above changes.

BI-5144


ch-tstrickland (commented):
There are a fair number of tabs in new code.

src/main/java/uk/gov/companieshouse/api/accounts/validation/TangibleAssetsValidator.java
```diff
@@ -57,14 +44,25 @@ public TangibleAssetsValidator(CompanyService companyService, CurrentPeriodServi
     private static final String NET_BOOK_VALUE_CURRENT_PERIOD = ".net_book_value_at_end_of_current_period";
     private static final String NET_BOOK_VALUE_PREVIOUS_PERIOD = ".net_book_value_at_end_of_previous_period";
 
+    private CurrentPeriodService currentPeriodService;
```
ch-tstrickland, line 47: Always make any member vars that can be final. It makes it clear to anyone reading that the field does not change and proofs your code against misguided maintenance.

ch-tstrickland (approved):
Talked through with Matt and all looks good to me.
final fields where possible.
Fix tabs.
Other than that, all looks great. Nice tidying up, consolidation.

Great Git commit comments. We should be doing more commits like that as a team.

PR #490, opened 2020-10-01
Fix logic too include current date

amoriartyCH:
-Added a day so that the logic would include the users current day as valid if within 7 of period_start_on

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/CompanyAccountValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/CompanyAccountValidator.java
@@ -29,7 +29,7 @@ import java.util.Optional;
 public class CompanyAccountValidator extends BaseValidator {
 
     private static final String ERROR_PATH = "$.company_account";
-    private static final int NUMBER_OF_DAYS = 7;
+    private static final int NUMBER_OF_DAYS = 8; // 7 days including the current date.
 
     @Autowired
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/CompanyAccountValidator.java
```diff
@@ -29,7 +29,7 @@
 public class CompanyAccountValidator extends BaseValidator {
 
     private static final String ERROR_PATH = "$.company_account";
-    private static final int NUMBER_OF_DAYS = 7;
+    private static final int NUMBER_OF_DAYS = 8; // 7 days including the current date.
```
mreis-ch, line 32: How did this come about?

amoriartyCH: With the 7 days originally this scenario would fail:

Period_start_on: 2020-10-08
Todays date: 2020-10-01

But we want too accept it on the 1st as thats technically 7 days before. So too make it work correctly i've added a day so that it counts the current day meaning the above scenario can now pass 👍

mreis-ch: Ah, right ok 👍

mreis-ch (approved):

PR #498, opened 2020-10-08
Remove unwanted if

faisal-amin82:
- Remove unnecessary if from the small-full-accounts template which causes issues with rendering ixbrl when balance sheet notes are not present

Resolves part of: BI-5443


mreis-ch (approved):

BSaunders95 (approved):

ch-tstrickland (approved):
Looks good.
Richer comment would be better.

PR #504, opened 2020-11-02
improve coverage and only give valid BS for prev period if multiyear …

amoriartyCH:
…filer = true

As commit states, we only want to give a balance sheet for previous period if we have a company which is infact a multi-year filer.

Previously we fed each note validator a new BS if they were a single year filer but I believe this is miss-leading, so updated to pass in null if previous period doesnt exist, and then in the stocks validator we do an Optional call to ensure the null exception would be handled.

Updated unit tests to improve coverage to 100% as well.

### DIFF
--- a/src/test/java/uk/gov/companieshouse/api/accounts/validation/AccountsValidatorTest.java
+++ b/src/test/java/uk/gov/companieshouse/api/accounts/validation/AccountsValidatorTest.java
@@ -137,6 +137,38 @@ class AccountsValidatorTest {
     }
 
     @Test
+    @DisplayName("Validate Submission - single year filer, successful, no errors")
+    void validateSubmissionSYFNoErrorsFound() throws DataException, ServiceException {
+
+        Errors emptyErrors = new Errors();
+
+        when(smallFullService.find(COMPANY_ACCOUNTS_ID, request)).thenReturn(smallFullResponseObject);
+        when(smallFullResponseObject.getData()).thenReturn(smallFull);
+
+        when(currentPeriodTnClosureValidator.validate(any(String.class), any(SmallFull.class),
+                any(HttpServletRequest.class), any(Errors.class))).thenReturn(emptyErrors);
+        when(previousPeriodTnClosureValidator.validate(any(String.class), any(SmallFull.class),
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/companieshouse/api/accounts/validation/AccountsValidatorTest.java
```diff
@@ -136,6 +136,38 @@ void setUp() throws DataException {
...
+                .thenReturn(emptyErrors);
+
+        when(currentPeriodService.find(COMPANY_ACCOUNTS_ID, request)).thenReturn(cpResponseObj);
+        when(cpResponseObj.getData()).thenReturn(currentPeriod);
+        when(currentPeriod.getBalanceSheet()).thenReturn(currentPeriodBalanceSheet);
+
+        when (companyService.isMultipleYearFiler(transaction)).thenReturn(false);
+
+        when(stocksTnClosureValidator
+                .validate(COMPANY_ACCOUNTS_ID, smallFull, transaction, request, emptyErrors, currentPeriodBalanceSheet, null))
```
ch-tstrickland, line 161: More readable if you use a local var with a meaningful name, assign that null, & pass it in. A bit verbose, but more readable.

amoriartyCH: good point, will add that now :)

ch-tstrickland (commented):
Leaving aside my very minor comment below, this looks good to me and a useful improvement.
I think that all of the PR text could usefully have gone in the commit message, which would be more useful to future maintainers.
Also the first line of a commit should be no more than 50 chars. You can tell when it's too long, as it gets truncated in a list of commits.

src/test/java/uk/gov/companieshouse/api/accounts/validation/AccountsValidatorTest.java
amoriartyCH: After tempting to add it, intellij recommended I change it back to null. Discussed and we will keep it null for now.

mreis-ch (approved):

PR #505, opened 2020-11-10
Update Spring Boot from 2.1.3 to 2.1.10

faisal-amin82:
This new version ensures to update the version of jackson-annotations dependency from 2.9.0 to 2.9.10
to ensure that Sonar does not complain about the unsafe data binding and unsafe deserialisation issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,9 +34,8 @@
     <commons-io.version>2.4</commons-io.version>
 
     <spring-boot-dependencies.version>2.1.10.RELEASE</spring-boot-dependencies.version>
-    <spring-boot-maven-plugin.version>2.1.3.RELEASE</spring-boot-maven-plugin.version>
+    <spring-boot-maven-plugin.version>2.1.10.RELEASE</spring-boot-maven-plugin.version>
 
-    <jackson.version>2.9.0.pr4</jackson.version>
     <!-- Maven and Surefire plugins -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -33,7 +33,7 @@
     <api-security-java.version>0.2.10</api-security-java.version>
     <commons-io.version>2.4</commons-io.version>
 
-    <spring-boot-dependencies.version>2.1.3.RELEASE</spring-boot-dependencies.version>
+    <spring-boot-dependencies.version>2.1.10.RELEASE</spring-boot-dependencies.version>
```
divya051988, line 36: Do you think we should also update the spring boot maven plugin version as well to keep inline with the spring boot dependency version upgrade ?

faisal-amin82: Updated spring boot maven version to match the dependencies version

divya051988 (approved):
Looks good to me but I noticed that we are on really old versions of structured logging and api-helper. These can be a potential security risk . Might be worth having a look with the team on how to upgrade the versions.

mreis-ch:
@divya051988 I've updated versions in the pom apart from structured logging as the latest version removes the json-20160810 library which is used by AccountsNoteFilter.class. Spoke with Tom and we'll do that as a separate task so it doesn't hold up this PR

divya051988 (approved):
Looks good worth running regression once just to be sure.

PR #506, opened 2020-11-11
Bump dependency

rafahop:
Upgrade `company-accounts-library` to latest version
Remove `api-security-java` which is not used and is causing version conflicts as it is already being imported by `company-accounts-library`


ch-tstrickland (approved):
Looks good to me.
For the purposes of future readers, the info that you put in the PR would have been better put in the commit messages.

PR #510, opened 2020-11-13
Add API service layer for RPT Transactions

mreis-ch:
Add models for both (Related Party Transactions) RPT Transaction entity and data
Add service and transformer classes for RPT Transaction

BI-5270

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
@@ -31,7 +31,8 @@ public enum Kind {
     LOANS_TO_DIRECTORS("small-full-accounts-note#loans-to-directors"),
     LOANS_TO_DIRECTORS_LOANS("small-full-accounts-loans-to-directors#loans"),
     LOANS_TO_DIRECTORS_ADDITIONAL_INFO("small-full-accounts-loans-to-directors#additional-information"),
-    RELATED_PARTY_TRANSACTIONS("small-full-accounts-note#related-party-transactions");
+    RELATED_PARTY_TRANSACTIONS("small-full-accounts-note#related-party-transactions"),
+    RPT_TRANSACTION("small-full-accounts-note#related-party-transactions-transaction");
 
     private String value;
 
--- a/src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
@@ -32,7 +32,7 @@ public enum Kind {
     LOANS_TO_DIRECTORS_LOANS("small-full-accounts-loans-to-directors#loans"),
     LOANS_TO_DIRECTORS_ADDITIONAL_INFO("small-full-accounts-loans-to-directors#additional-information"),
     RELATED_PARTY_TRANSACTIONS("small-full-accounts-note#related-party-transactions"),
-    RPT_TRANSACTION("small-full-accounts-note#related-party-transactions-transaction");
+    RPT_TRANSACTIONS("small-full-accounts-note-related-party-transactions#transactions");
 
     private String value;
 
--- a/src/main/java/uk/gov/companieshouse/api/accounts/ResourceName.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/ResourceName.java
@@ -28,6 +28,7 @@ public enum ResourceName {
     LOANS_TO_DIRECTORS("loans-to-directors"),
     LOANS("loans"),
     RELATED_PARTY_TRANSACTIONS("related-party-transactions"),
+    RPT_TRANSACTIONS("related-party-transactions-transactions"),
     ADDITIONAL_INFO("additional-information");
 
     private String name;
--- a/src/main/java/uk/gov/companieshouse/api/accounts/ResourceName.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/ResourceName.java
@@ -28,7 +28,7 @@ public enum ResourceName {
     LOANS_TO_DIRECTORS("loans-to-directors"),
     LOANS("loans"),
     RELATED_PARTY_TRANSACTIONS("related-party-transactions"),
-    RPT_TRANSACTIONS("related-party-transactions-transactions"),
+    RPT_TRANSACTIONS("transactions"),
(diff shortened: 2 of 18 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/Kind.java
```diff
@@ -31,7 +31,8 @@
     LOANS_TO_DIRECTORS("small-full-accounts-note#loans-to-directors"),
     LOANS_TO_DIRECTORS_LOANS("small-full-accounts-loans-to-directors#loans"),
     LOANS_TO_DIRECTORS_ADDITIONAL_INFO("small-full-accounts-loans-to-directors#additional-information"),
-    RELATED_PARTY_TRANSACTIONS("small-full-accounts-note#related-party-transactions");
+    RELATED_PARTY_TRANSACTIONS("small-full-accounts-note#related-party-transactions"),
+    RPT_TRANSACTION("small-full-accounts-note#related-party-transactions-transaction");
```
amoriartyCH, line 35: I believe this should be:
RPT_TRANSACTIONS("small-full-accounts-related-party-transactions#transactions")

src/main/java/uk/gov/companieshouse/api/accounts/ResourceName.java
```diff
@@ -28,6 +28,7 @@
     LOANS_TO_DIRECTORS("loans-to-directors"),
     LOANS("loans"),
     RELATED_PARTY_TRANSACTIONS("related-party-transactions"),
+    RPT_TRANSACTIONS("related-party-transactions-transactions"),
```
amoriartyCH, line 31: I believe this one should be similar to the LOANS above, so something like:
RPT_TRANSACTIONS("transactions")

amoriartyCH (commented):
awesome stuff, minor comments :)

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/smallfull/notes/relatedpartytransactions/RptTransactionEntity.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.companieshouse.api.accounts.model.entity.smallfull.notes.relatedpartytransactions;
+
+import org.springframework.data.mongodb.core.mapping.Field;
+import uk.gov.companieshouse.api.accounts.model.entity.BaseEntity;
+
+public class RptTransactionEntity extends BaseEntity {
```
faisal-amin82, line 6: I believe we need `@Document` annotation here with the collection name notes otherwise it will create another collection with the class name

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/RptTransactionServiceImpl.java
```diff
@@ -0,0 +1,216 @@
...
+    private static final String RPT_TRANSACTION_LINK = "loans";
+
+    @Autowired
+    private RptTransactionTransformer transformer;
+
+    @Autowired
+    private RptTransactionRepository repository;
+
+    @Autowired
+    private LoansToDirectorsServiceImpl loansToDirectorsService;
```
faisal-amin82, line 41: I think this would be the RPT parent resource?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/RptTransactionServiceImpl.java
```diff
@@ -155,8 +161,6 @@
 
                 repository.deleteById(rptTransactionId);
```
amoriartyCH, line 163: When we delete a transaction we will need to remove it from the parent resource, similar to LTD loans delete functionality

src/main/java/uk/gov/companieshouse/api/accounts/model/entity/smallfull/notes/relatedpartytransactions/RptTransactionEntity.java
faisal-amin82: I think it would go inside the notes collection as this is a note (`RptTransaction`) and not a `Transaction`

amoriartyCH:
Is it possible for the test coverage to be upped to closer to the usual 95% before we merge this PR (discussed in standup so noting here as Matt wasn't available during this time)

mreis-ch:
Will look to eek out another few % of unit test coverage 👍

src/main/java/uk/gov/companieshouse/api/accounts/service/RptTransactionService.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.companieshouse.api.accounts.service;
+
+public interface RptTransactionService {
```
amoriartyCH, line 3: Just noticed this, I believe we have a "MultipleResourceService<T>" which we should implement, it comes with a findAll() and deleteAll() method. So could do with deleting this service and then in the RptTransactionServiceImpl we can do "implements MultipleResourceService<RptTransaction>

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/RelatedPartyTransactionsServiceImpl.java
```diff
@@ -142,6 +141,69 @@ public void removeLink(String id, RelatedPartyTransactionsLinkType linkType, Htt
...
+    }
+
+    @Override
+    public void removeRptTransaction(String companyAccountsId, String rptTransactionId, HttpServletRequest request)
+            throws DataException {
+
+        String resourceId = generateID(companyAccountsId);
+        RelatedPartyTransactionsEntity entity = repository.findById(resourceId)
+                .orElseThrow(() -> new DataException(
+                        "Failed to find loans to directors entity from which to remove loan"));
```
amoriartyCH, line 174: the data exception message should be updated

src/main/java/uk/gov/companieshouse/api/accounts/service/RptTransactionService.java
mreis-ch: Already implement the MultipleResourceService so will just remove the interface 👍

amoriartyCH (approved):

PR #511, opened 2020-11-13
RPT Transaction First Year Filer Validation

faisal-amin82:
- Adds validation related code where if user is a single year filer and provides data for breakdowns' balance at period start then it would throw an error stating that the data is not expected 

Resolves: BI-5281


src/main/java/uk/gov/companieshouse/api/accounts/model/rest/smallfull/notes/relatedpartytransactions/RptTransaction.java
```diff
@@ -3,22 +3,27 @@
...
     @JsonProperty("name_of_related_party")
     private String nameOfRelatedParty;
 
+    @NotNull
     @JsonProperty("relationship")
     private String relationship;
 
+    @NotNull
     @JsonProperty("description_of_transaction")
     private String description;
```
amoriartyCH, line 20: will this cause issues with de-serialising? usually we match the name of the json property to the java variable just in camel case, can we review this?

PR #516, opened 2020-11-18
Add character/numeric validation

mreis-ch:
Add invalid character and character length validation annotations to RptTransaction
BI-5272

Add number range validation annotations to RptTransactionBreakdown
BI-5273


ch-tstrickland (commented):
I know that coverage does not capture this, but should we be unit testing these validations?

mreis-ch:
> I know that coverage does not capture this, but should we be unit testing these validations?

It's a fair point. I'll take a look

ch-tstrickland (approved):

PR #520, opened 2020-11-24
Fix sonar code smell issues

mreis-ch:
- Rename class members that have the same name as their classes Costs, Error and Errors. Update getters and setters and classes that call these.
- Reduce verbosity of generics in AccountsNoteConverter
- Use primitive boolean in FilingServiceImpl
- The majority of changes are the removal of the public modifier from unit tests and methods within.
- Removed any unused imports or variables which were not used.
- Added, where possible, parameterised generics

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
@@ -6,14 +6,14 @@ import java.util.Map;
 public class Costs {
 
     @JsonProperty("costs")
-    private Map<String, Cost> costsList;
+    private Map<String, Cost> costsMap;
 
     public Map<String, Cost> getCostsList() {
-        return costsList;
+        return costsMap;
     }
 
     public void setCostsList(
             Map<String, Cost> costs) {
-        this.costsList = costs;
+        this.costsMap = costs;
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
@@ -6,14 +6,14 @@ import java.util.Map;
 public class Costs {
 
     @JsonProperty("costs")
-    private Map<String, Cost> costs;
+    private Map<String, Cost> costsList;
 
-    public Map<String, Cost> getCosts() {
-        return costs;
+    public Map<String, Cost> getCostsList() {
+        return costsList;
     }
 
-    public void setCosts(
+    public void setCostsList(
             Map<String, Cost> costs) {
-        this.costs = costs;
+        this.costsList = costs;
     }
 }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
@@ -8,11 +8,11 @@ public class Costs {
     @JsonProperty("costs")
     private Map<String, Cost> costsMap;
 
-    public Map<String, Cost> getCostsList() {
+    public Map<String, Cost> getCostsMap() {
         return costsMap;
     }
 
-    public void setCostsList(
+    public void setCostsMap(
             Map<String, Cost> costs) {
         this.costsMap = costs;
     }
--- a/src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Errors.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Errors.java
@@ -13,7 +13,7 @@ public class Errors {
 
     @JsonProperty("errors")
     @Field("errors")
-    private Set<Error> errorsList = new HashSet<>();
+    private Set<Error> errorsSet = new HashSet<>();
 
     /**
      * Add the given {@link Error}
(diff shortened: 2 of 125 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Errors.java
```diff
@@ -13,7 +13,7 @@
 
     @JsonProperty("errors")
     @Field("errors")
-    private Set<Error> errors = new HashSet<>();
+    private Set<Error> errorsList = new HashSet<>();
```
amoriartyCH, line 16: Same as the previous comment, could this cause issues with spring?

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
```diff
@@ -6,14 +6,14 @@
 public class Costs {
 
     @JsonProperty("costs")
-    private Map<String, Cost> costs;
+    private Map<String, Cost> costsList;
```
amoriartyCH, line 9: Can this cause issue with spring-deserialising? usually the name of the field has to match the json property although i could be wrong (worth checking this)

src/test/java/uk/gov/companieshouse/api/accounts/ApplicationConfigurationTest.java
```diff
@@ -9,25 +9,25 @@
 
 import uk.gov.companieshouse.environment.EnvironmentReader;
 
-public class ApplicationConfigurationTest {
+class ApplicationConfigurationTest {
```
amoriartyCH, line 12: Do we know the default if public isn't specified? (Was this private before?)

amoriartyCH (commented):
Looks good, before approving wanted to ask if we are okay with all of these changes going into 1 very large PR? As this will involve a lot of different testing scenarios for one ticket without the ability to test each change in isolation?

src/test/java/uk/gov/companieshouse/api/accounts/ApplicationConfigurationTest.java
mreis-ch: From what I can see it's package-private

amoriartyCH: awesome

amoriartyCH (approved):

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
ch-tstrickland: The point of JsonProperty is to rename a property.
https://github.com/FasterXML/jackson-annotations#annotations-for-renaming-properties

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Errors.java
ch-tstrickland: I don't think so. See previous answer.

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
amoriartyCH: yeah, investigated this and talked about it in slack, all good 👍

ch-tstrickland, line 9: Would costsMap be better here? It is not a list.

src/main/java/uk/gov/companieshouse/api/accounts/model/validation/Errors.java
ch-tstrickland: errorSet?

src/test/java/uk/gov/companieshouse/api/accounts/ApplicationConfigurationTest.java
ch-tstrickland: Yes, package private by default if unspecified in Java.

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
mreis-ch: That's a fair point. I'll update

ch-tstrickland (commented):
Minor comments aside, looks good to me.

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
```diff
@@ -6,14 +6,14 @@
...
+    private Map<String, Cost> costsMap;
 
-    public Map<String, Cost> getCosts() {
-        return costs;
+    public Map<String, Cost> getCostsList() {
+        return costsMap;
     }
 
-    public void setCosts(
+    public void setCostsList(
```
ch-tstrickland, line 15: setCostsMap

src/main/java/uk/gov/companieshouse/api/accounts/model/rest/Costs.java
```diff
@@ -6,14 +6,14 @@
 public class Costs {
 
     @JsonProperty("costs")
-    private Map<String, Cost> costs;
+    private Map<String, Cost> costsMap;
 
-    public Map<String, Cost> getCosts() {
-        return costs;
+    public Map<String, Cost> getCostsList() {
```
ch-tstrickland, line 11: getCostsMap

ch-tstrickland (approved):

PR #524, opened 2020-12-02
Add validation for RPT transactions, transaction type

faisal-amin82:
This commit checks whether the value supplied for transaction type field of transactions is one of the two acceptable values.
If the value supplied does not match either of the two sentences then it throws invalid data error

Resolves: BI-6264

image

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
@@ -12,8 +12,6 @@ import uk.gov.companieshouse.api.model.transaction.Transaction;
 public class RptTransactionValidator extends BaseValidator {
 
     private static final String TRANSACTION_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START = "$.transaction.breakdown.balance_at_period_start";
-    private static final String[] TRANSACTION_TYPE_MESSAGE = {"money given to the company by a related party", "money given to a related party by the company"};
-    private static final String TRANSACTION_TYPE = "$.transaction.transaction_type";
 
     @Autowired
     public RptTransactionValidator(CompanyService companyService) {
@@ -31,12 +29,6 @@ public class RptTransactionValidator extends BaseValidator {
             return errors;
         } else if (isMultipleYearFiler && rptTransaction.getBreakdown().getBalanceAtPeriodStart() == null) {
             addError(errors, mandatoryElementMissing, TRANSACTION_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START);
-            return errors;
-        }
-
-        if (!((rptTransaction.getTransactionType().trim().equalsIgnoreCase(TRANSACTION_TYPE_MESSAGE[0]) ||
-                rptTransaction.getTransactionType().trim().equalsIgnoreCase(TRANSACTION_TYPE_MESSAGE[1])))) {
-            addError(errors, invalidValue, TRANSACTION_TYPE);
             return errors;
         }
 
--- a/src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
@@ -12,6 +12,8 @@ import uk.gov.companieshouse.api.model.transaction.Transaction;
 public class RptTransactionValidator extends BaseValidator {
 
     private static final String TRANSACTION_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START = "$.transaction.breakdown.balance_at_period_start";
+    private static final String[] TRANSACTION_TYPE_MESSAGE = {"money given to the company by a related party", "money given to a related party by the company"};
+    private static final String TRANSACTION_TYPE = "$.transaction.transaction_type";
 
     @Autowired
     public RptTransactionValidator(CompanyService companyService) {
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
```diff
@@ -12,6 +12,8 @@
 public class RptTransactionValidator extends BaseValidator {
 
     private static final String TRANSACTION_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START = "$.transaction.breakdown.balance_at_period_start";
+    private static final String[] TRANSACTION_TYPE_MESSAGE = {"money given to the company by a related party", "money given to a related party by the company"};
```
mreis-ch, line 15: Why use an array here? Wouldn't an Map be better and use the .containsValue method to check if the correct value or an Enum and use the EnumUtils to check?
Have we decided to send the whole message from the web or a key that is stored?

src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
```diff
@@ -24,14 +26,20 @@ public Errors validateRptTransaction(RptTransaction rptTransaction, Transaction
...
-        if(!isMultipleYearFiler && rptTransaction.getBreakdown().getBalanceAtPeriodStart() != null ) {
+        if (!isMultipleYearFiler && rptTransaction.getBreakdown().getBalanceAtPeriodStart() != null) {
             addError(errors, unexpectedData, TRANSACTION_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START);
             return errors;
         } else if (isMultipleYearFiler && rptTransaction.getBreakdown().getBalanceAtPeriodStart() == null) {
             addError(errors, mandatoryElementMissing, TRANSACTION_BREAKDOWN_PATH_BALANCE_AT_PERIOD_START);
             return errors;
         }
 
+        if (!((rptTransaction.getTransactionType().trim().equalsIgnoreCase(TRANSACTION_TYPE_MESSAGE[0]) ||
```
mreis-ch, line 37: As previous comment, if a Map used then can use the .containsValue method or with an Enum, the EnumUtils.isValidEnum method if we're only storing key in DB

src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
amoriartyCH: I believe with FRS101/102 we send the string as a whole to the API from the WEB. As for using array i'd say it does seem strange but would work fine, could even simply it and take them out of the array and just have them as 2 constants? See what everyone else says 👍

src/main/java/uk/gov/companieshouse/api/accounts/validation/RptTransactionValidator.java
ch-tstrickland: Which EnumUtils is that? I cannot find it.
Anyway, I agree with the idea of mapping to an enum, given a choice.
I don't think that maps are more efficient than a short list for very small collections.
Although something like Google's immutable map might be more efficient than many, you still end up calculating a hash for the string to look it up.

ch-tstrickland: "I still vividly remember the day that Wirth decided to replace the elegant data structure used in the compiler’s symbol table handler by a mundane linear list. In the original compiler, the objects in the symbol table had been sorted in a tree data structure (in identifier lexical order) for fast access, with a separate linear list representing their declaration order. One day Wirth decided that there really weren’t enough objects in a typical scope to make the sorted tree cost-effective. All of us Ph.D. students were horrified: it had taken time to implement the sorted tree, the solution was elegant, and it worked well – so why would one want to throw it away and replace it by something simpler, and even worse, something as prosaic as a linear list? But of course, Wirth was right, and the simplified compiler was both smaller and faster than its predecessor."
http://www.ics.uci.edu/~franz/Site/pubs-pdf/BC03.pdf

src/main/java/uk/gov/companieshouse/api/accounts/validation/ValidRptTransactionTypeImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+@Component
+public class ValidRptTransactionTypeImpl implements ConstraintValidator<ValidRptTransactionType, String> {
+
+    @Override
+    public boolean isValid(String rptTransactionType, ConstraintValidatorContext context) {
+        return StringUtils.isBlank(rptTransactionType)
+                || rptTransactionType.trim().equalsIgnoreCase("money given to the company by a related party")
```
ch-tstrickland, line 15: Do we want to use ignorecase here?
What do we do elsewhere?
I only ask because these are machine-supplied values, not user-input. In a public facing API, I would make these enums with exact values. All of which means that you could tightly specify the string, including its case.

ch-tstrickland: Regarding trim...
Do you think that we should trim user input and store the trimmed form?

src/test/java/uk/gov/companieshouse/api/accounts/validation/ValidRptTransactionTypeImplTest.java
```diff
@@ -0,0 +1,62 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ExtendWith(MockitoExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class ValidRptTransactionTypeImplTest {
+
+    @Mock
+    private ConstraintValidatorContext context;
+
+    private static final String MONEY_GIVEN_TO_RELATED_PARTY = "Money given to a related party by the company";
```
ch-tstrickland, line 23: move constants to top

faisal-amin82: Moved constants to the top

src/main/java/uk/gov/companieshouse/api/accounts/validation/ValidRptTransactionTypeImpl.java
faisal-amin82: We are trimming and are allowing when it comes to specifying the string in mixed case in for example FRS101 and FRS102 case. I remember, you i and Brad had discussion about this regarding FRS101 and FRS102 and we came up with trimming and ignoring cases when it comes to compare the user input with our values 🙂 

I am more than happy to use enums and restrict the user from providing exact value but it might make it a bit more difficult in case they have a habit of typing capitals for example

mreis-ch: The user can't provide an exact value can they? The text is from the radio button which we're supplying so we might be ok with enums with this?

faisal-amin82: @mreis-ch This will be for API filers who use for example Postman to file their accounts. For web filers the value will be same as one of the selected radios

ch-tstrickland (approved):

PR #526, opened 2020-12-17
Resolves issues with mandatory element missing on stocks note

faisal-amin82:
This commit resolves issues where users when they enter 0 in the figure for stocks on balance sheet. When users enter 0 or leave the field blank on balance sheet then it does not display the note in the journey. Transactions closure validation previously looked into this and based on the figure it validated the value. This commit in addition to transaction closure resolves the issue

Resolves: BI-6485


ch-tstrickland (commented):
So the fix is to not leave it unset anymore, but:
1. to set it to zero if it's missing in either current or previous. It will thus be zero either because it was set to that by the user, or because the local var was set to that by being null
2. if it is not zero, then error.

Is this all correct?

I am puzzled that this has not resulted in any unit test changes.

ch-tstrickland:
I have spoken with Faisal, who is adding one unit test to cover the new paths (if someone sets either value to zero, then it should not be considered an error). With that in place, I approve. If I am not present and those tests are in, could someone else approve please?

amoriartyCH (approved):

PR #528, opened 2021-03-24
Bump company accounts library

rafahop:
The new version no longer uses the enable token permissions auth feature flag

Resolves FA-1070


CHdlewis (approved):
Looks ok

PR #541, opened 2021-12-15
Update log4j to safe version

rmcmurray-ch:
Updates log4j to version 2.16.0 and structured-logging to version 1.9.11.

This dependency has been directly added to the pom.xml file due to springboot dependency management bringing in old version of log4j regardless of what is being  used in structured-logging. This will should be removed when springboot has been upgraded to the required patch 2.5.8 or  
2.6.2.
More information can be found here: https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot
Structured-logging version updated to 1.9.10.

Resolved issue of overloaded methods in structured logging library.

Resolves DEBT-1436.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,6 +26,7 @@
     <!-- Dependency Versions -->
     <api-helper-java.version>1.4.3</api-helper-java.version>
     <structured-logging.version>1.9.11</structured-logging.version>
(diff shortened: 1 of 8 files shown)


CHdlewis (approved):
LGTM

PR #548, opened 2022-12-12
BI-12238-Remove-addToLog-from-TnepValidationServiceImpl

OsvaldoMartini:
##### Removed AddToLog from Tnep Validation Service

### DIFF
--- a/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
+++ b/src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
@@ -55,19 +55,10 @@ public class TnepValidationServiceImpl implements TnepValidationService {
             Results results = validatIxbrlAgainstTnep(ixbrl, location);
 
             if (hasPassedTnepValidation(results)) {
-                addToLog(false, null, results, location,
-                    "Ixbrl is valid. It has passed the TNEP validation");
-
                 isIxbrlValid = true;
-
-            } else {
-                addToLog(true, null, results, location,
-                    "Ixbrl is invalid. It has failed the TNEP validation");
             }
 
         } catch (Exception e) {
-            addToLog(true, e, null, location,
-                "Exception has been thrown when calling TNEP validator. Unable to validate Ixbrl");
         }
 
         LOGGER.info("TnepValidationServiceImpl: Ixbrl validation has finished");
@@ -107,24 +98,6 @@ public class TnepValidationServiceImpl implements TnepValidationService {
 
         return restTemplate
             .postForObject(new URI(getIxbrlValidatorUri()), requestEntity, Results.class);
-    }
-
-    private void addToLog(boolean hasValidationFailed, Exception e, Results results,
-        String location, String message) {
-
-        Map<String, Object> logMap = new HashMap<>();
-        logMap.put("message", message);
-        logMap.put("location", location);
-
-        if (results != null) {
-            logMap.put("results", results);
-        }
-
-        if (hasValidationFailed) {
-            LOGGER.error("TnepValidationServiceImpl: validation has failed", e, logMap);
-        } else {
-            LOGGER.debug("TnepValidationServiceImpl: validation has passed", logMap);
-        }
     }
 
     /**


src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
```diff
@@ -55,19 +55,10 @@ public boolean validate(String ixbrl, String location) {
...
-                    "Ixbrl is valid. It has passed the TNEP validation");
-
                 isIxbrlValid = true;
-
-            } else {
-                addToLog(true, null, results, location,
-                    "Ixbrl is invalid. It has failed the TNEP validation");
             }
 
         } catch (Exception e) {
```
HarryBosworth, line 61: While we removed the addToLog call, I think it would be a good idea to still include an error log in the case of an exception.

ch-tstrickland (commented):
Are we sure this is the best resolution of this bug? It seems to remove ALL logging.
I wonder if the issue at hand has been properly specified?

The issue: logging "results" is probably wrong. It contains customer-specific data.

ch-tstrickland:
Is there a ticket associated with this? What is its number? That might help to better specify the fix for you.

ch-tstrickland:
Ah, sorry - I see it in the branch name, though it is convention also to put it in the PR's description.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
```diff
@@ -55,19 +55,10 @@ public boolean validate(String ixbrl, String location) {
...
-            addToLog(true, e, null, location,
-                "Exception has been thrown when calling TNEP validator. Unable to validate Ixbrl");
         }
 
         LOGGER.info("TnepValidationServiceImpl: Ixbrl validation has finished");
```
JMetherell, line 64: I'd suggest updating this log line to include whether the validation was successful.

ch-tstrickland:
There may even be useful data in the results object.
Looking at:
https://github.com/companieshouse/company-accounts.api.ch.gov.uk/pull/548/files#diff-92c5ecc9906d3150ce313fa0b1f7a1aabe50be11c75a4d8aceebca0c5d40d322R87

There is info in there about pass/fail. There may be other useful info. Look at:
https://companieshouse.slack.com/archives/C04ET2Y90HG/p1670857289913319

The Results object looks like this:
https://github.com/companieshouse/company-accounts.api.ch.gov.uk/blob/master/src/main/java/uk/gov/companieshouse/api/accounts/validation/Results.java

So, to my mind, in the event of an error, we should be fishing out the content of Result.getValidationStatus and Result.getErrors and logging that. The real problem lies in the data field of Result.

I also worry that Result has a toString on it that includes the data, which looks like a general risk.

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
```diff
@@ -55,19 +55,10 @@ public boolean validate(String ixbrl, String location) {
...
-
                 isIxbrlValid = true;
-
-            } else {
-                addToLog(true, null, results, location,
-                    "Ixbrl is invalid. It has failed the TNEP validation");
             }
 
         } catch (Exception e) {
-            addToLog(true, e, null, location,
```
Jata2k, line 69: We shouldn't just be swallowing the exception, we should still log it

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
```diff
@@ -55,19 +55,10 @@ public boolean validate(String ixbrl, String location) {
             Results results = validatIxbrlAgainstTnep(ixbrl, location);
 
             if (hasPassedTnepValidation(results)) {
-                addToLog(false, null, results, location,
```
Jata2k, line 58: We should still log this, we only need omit the entire results object from the log. That's what's causing the logging issues

OsvaldoMartini:
> Are we sure this is the best resolution of this bug? It seems to remove ALL logging. I wonder if the issue at hand has been properly specified?
> 
> The issue: logging "results" is probably wrong. It contains customer-specific data.

I rolled back and removed only "results" from the **addToLog**

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
OsvaldoMartini: I rolled back and removed only "results" from the **addToLog**

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
OsvaldoMartini: I rolled back and removed only "results" from the **addToLog**

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
OsvaldoMartini: I rolled back and removed only "results" from the **addToLog**

OsvaldoMartini:
> 

will do it

ch-tstrickland (commented):
This looks good. My only question is whether it is worth including the Result.getValidationStatus text(), but since I can't see what's in it other than OK, I had a look around and instead have a more general Q, which I will raise separately to this PR which should remain tight. My Q:
@JMetherell   @Jata2k Do these logs need any extra value(s) for context, or are they enough? Do they need something like a requestId or similar to help with correlation?

src/main/java/uk/gov/companieshouse/api/accounts/service/impl/TnepValidationServiceImpl.java
```diff
@@ -133,16 +128,16 @@ private void addToLog(boolean hasValidationFailed, Exception e, Results results,
...
         headers.setContentType(MediaType.MULTIPART_FORM_DATA);
 
         return new HttpEntity<>(map,
-            headers);
+                headers);
     }
 
     private LinkedMultiValueMap<String, Object> createFileMessageResource(String ixbrl,
-        String location) {
+                                                                          String location) {
```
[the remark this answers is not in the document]
OsvaldoMartini: I will roll back to the original and block the ide auto-format

ch-tstrickland (approved):

PR #562, opened 2024-04-30
update log level info

tmoller-ch:


terraform/groups/ecs-service/locals.tf
gjackson-ch, line 1: Extra comma may cause issues

gjackson-ch (approved):

PR #578, opened 2024-07-08
Role for s3 access

amartin7663:
This pr creates a new role for the service to allow it access to the document renderer bucket so that it can access the accounts image at the end of the journey ready to be sent to CHIPS.


ch-code-analysis:
CI: No IaC security warnings found

chillier-ch (approved):

PR #579, opened 2024-07-08
add jaxb-impl dependency to fix xml desreialisation error on validation

amartin7663:
This pr adds the jaxp-impl dependency, to fix an issue caused by the failure to read the xml result returned by the felix validator when the accounts are sent to be validated.


ch-code-analysis:
CI: Security warnings found!

hwatkins-ch (approved):

